Issue #54, opened 2026-03-12
[Feature]: Go Producer Client Library for Async Queue API


RishabhSaini:
### Problem Statement

 No client library exists for submitting requests and retrieving results from the async processor. Clients must manually interact with queue-specific APIs (Redis, GCP Pub/Sub), handle serialization, and manage result correlation.

 ### Proposed Solution
 Create abstract `Producer` interface with queue-agnostic API:
 ```go
 SubmitRequest(ctx, Request) error
 GetResult(ctx, requestID) (*Result, error)
 ```

 Implementations: `RedisSortedSetProducer`, `RedisPubSubProducer`, `GCPPubSubProducer`

 ## Multi-Producer Result Correlation
 Challenge 1: Multiple producers must receive only their own results without race conditions.
Challenge 2 : Should FIFO queues support getting result by requestIDs?

 **Solutions:**
 1. **Per-client result queues**: Each producer instance routes results to unique queue via request metadata

### Willingness to Contribute

Yes, I can submit a PR

evacchi:
personally the way I was thinking about this was basically 1 dispatcher (i.e. llm-d-async instance) = 1 EPP

PR #55, opened 2026-03-12
feat: add Go producer client library for async queue API

RishabhSaini:
Resolves: #54 
  
Implement Producer interface with Redis sorted set backend for submitting
  requests and retrieving results from async processing queue.

  - Add Producer interface using existing api.RequestMessage/ResultMessage types
  - Implement RedisSortedSetProducer with multi-tenant isolation via TenantID
  - Results namespaced as results:{tenantID}:{queueName} to prevent collisions
  - Update sortedset_impl resultWorker to route results via metadata
  - Add comprehensive tests covering validation, error handling, and tenant isolation

  Multi-tenant safe: each tenant gets isolated result queue while sharing
  request queue for efficient priority-based processing.

### DIFF
--- a/pkg/producer/redis_sortedset_producer.go
+++ b/pkg/producer/redis_sortedset_producer.go
@@ -1,0 +1,215 @@
+package producer
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/llm-d-incubation/llm-d-async/pkg/async/api"
+	"github.com/redis/go-redis/v9"
+)
+
+// RedisSortedSetProducer implements Producer using Redis sorted set for requests
+// and Redis list for results.
+type RedisSortedSetProducer struct {
+	client           *redis.Client
+	requestQueueName string
+	resultQueueName  string
+}
+
+// RedisSortedSetConfig contains configuration for the Redis sorted set producer.
+type RedisSortedSetConfig struct {
+	// RedisAddr is the Redis server address (e.g., "localhost:6379").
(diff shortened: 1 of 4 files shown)


RishabhSaini:
Current known limitations of this approach: 
1) No request-result correlation. The resonsibility of tracking request IDs falls on the caller
2) Does not support batch addition/removal into/from the queue

pkg/producer/redis_sortedset_producer.go
```diff
@@ -0,0 +1,215 @@
...
+		return nil, errors.New("TenantID is required for multi-tenant isolation")
+	}
+
+	if config.RequestQueueName == "" {
+		config.RequestQueueName = "request-sortedset"
+	}
+
+	if config.ResultQueueName == "" {
+		config.ResultQueueName = "default"
+	}
```
evacchi: relatively minor issue, but we might want to ensure that TenantID and ResultQueueName do not contain `:` otherwise the namespaced name might be ambiguous

pkg/producer/redis_sortedset_producer.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return &result, nil
+}
+
+// Close closes the Redis connection.
+func (p *RedisSortedSetProducer) Close() error {
+	return p.client.Close()
+}
+
+// QueueDepth returns the number of pending requests in the queue.
```
evacchi: note: this method (and the others below) are currently not part of the interface, and at this time they aren't referenced anywhere else. It doesn't mean they have to be removed, just FYI

PR #60, opened 2026-03-16
feat: add metrics interface, saturation metric dispatch gate, e2e tests

### LINKED ISSUE #51: [Feature]: Metrics consumer
### Problem Statement

An immediate use case for the dispatch gate is to compute saturation metrics depending on some metrics. The consumer of such metrics should be an abstract interface that can be implemented in different ways both for testability and flexibility

### Proposed Solution

Define and provide an implementation for a metrics consumer modeled after GAIE/LLM-D. Ideally the consumer might read from prometheus or even scrape directly from VLLM if low-latency access to the metrics is needed (and if feasible)

### Alternatives Considered

- the implementation does not necessarily need an interface: the two could be developed independently and eventually we could define an adapter -- personally, I would follow whatever was done in GAIE, possibly even reusing and depending on parts of it if it makes sense
- instead of deploying llm-d-async separately, it could be a library and LLM-D could depend on it; in that case it could access the same metrics directly (but it would be still reading metrics)
- instead of reading metrics, alternative gating strategies could be devised -- these can be still provided as alternative implementation of the dispatch gate interface

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

evacchi:
## What does this PR do?

- Introduces the `MetricSource` interface that allows monitoring several labels at once
- provides a Prometheus implementation that maps the requested labels to Prom queries
- use the Prom API client instead of raw HTTP requests
- refactors the BinaryMetric gate to use it 
- adds test cases to BinaryMetric validating the new impl follows the original impl
- introduces a DispatchGate for the saturation metric in GAIE
- add unit and e2e tests (e2e use a mock Prometheus for now)

## Why is this change needed?

Continuation of #2, work on #51.

## How was this tested?

## Checklist

## Related Issues

Closes #51

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -245,8 +245,7 @@ func setupRedisClient() {
 		if err != nil {
 			return err
 		}
-		resp.Body.Close()
-		return nil
+		return resp.Body.Close()
 	}, 30*time.Second, 1*time.Second).Should(gomega.Succeed())
 }
 


pkg/async/inference/flowcontrol/binary_metric_dispatch_gate.go
```diff
@@ -35,105 +30,67 @@ var prometheusURL = flag.String("gate.prometheus.url", "", "Prometheus URL for n
...
-		logger.V(logutil.DEFAULT).Info("Expected Vector result, got: ", result)
+		logger.V(logutil.DEFAULT).Info("MetricSource error, failing open", "error", err)
 		return 1.0
 	}
 
-	if len(vec) == 0 {
-		logger.V(logutil.DEFAULT).Info("No metrics found for the given query.")
+	if len(samples) == 0 {
+		logger.V(logutil.DEFAULT).Info("No metrics found, failing open")
 		return 1.0
```
RishabhSaini: In the case the metric is missing/unavailable:
budget = 1.0 for BinaryMetricDispatchGate
budget = 0.0 for SaturationMetricDispatchGate

We need consistent behaviour here

PR #84, opened 2026-03-27
added gcp pubsub to e2e tests

shimib:
## What does this PR do?

Added e2e test support for GCP PubSub based queue implementation. Using the PubSub emulator.

## How was this tested?

- [X ] Integration/e2e tests added/updated

## Checklist

## Related Issues
https://github.com/llm-d-incubation/llm-d-async/issues/83

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
@@ -12,6 +13,9 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
@@ -77,6 +81,8 @@ var (
 	pubsubClient *pubsub.Client
 	adminURL     string
 	promMockURL  string
+
+	usedPorts = make(map[string]bool)
 )
 
 func TestEndToEnd(t *testing.T) {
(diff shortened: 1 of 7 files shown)


test/e2e/yaml/pubsub-emulator.yaml
```diff
@@ -0,0 +1,46 @@
...
+  name: pubsub-emulator
+  namespace: e2e-test
+spec:
+  type: NodePort
+  selector:
+    app: pubsub-emulator
+  ports:
+    - port: 8085
+      targetPort: 8085
+      nodePort: 30850
```
crissepulveda-0: Hardcoding a NodePort here creates a high risk of port collisions if multiple E2E tests run on the same node or if the environment has other services in this range. While  internal container ports are fine and particular line matches your pubsubPort default in the Go code, NodePorts are a global host resource. Consider if this can be dynamically assigned or if there's a guard to ensure this port is free before starting the suite.

test/e2e/e2e_suite_test.go
```diff
@@ -249,6 +282,54 @@ func setupRedisClient() {
...
+	os.Setenv("PUBSUB_EMULATOR_HOST", "localhost:"+pubsubPort) // nolint:errcheck
+
+	var err error
+	pubsubClient, err = pubsub.NewClient(ctx, "test-project")
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	// Create request and result topics/subscriptions
+	ginkgo.By("Creating request-topic and request-sub")
+	reqTopic, err := pubsubClient.CreateTopic(ctx, "request-topic")
+	if err != nil && !strings.Contains(err.Error(), "AlreadyExists") {
```
crissepulveda-0: This line (and the other `if err != nil && !strings.Contains(err.Error(), "AlreadyExists")` lines are good to check but a bit brittle.  Was looking into this a bit (and consulted Gemini so take this with a grain of salt...BUT) the pubsub SDK seems to mainly return specific GRPC error codes. Using something that checks the error code like `status.Code(err) == codes.AlreadyExists` would be more robust.

crissepulveda-0 (changes requested):

PR #115, opened 2026-04-17
tests: remove mock e2e tests, use llm-d-inference-sim and real prometheus

evacchi:
required helm charts + doc changes spun off to #163 

## What does this PR do?

  Replaces the hand-rolled mock services (`igw-mock`, `prom-mock`) in the e2e test suite with  [llm-d-inference-sim](https://github.com/llm-d/llm-d-inference-sim), an actual Envoy instance (with fault injection to simulate 5xx errors), the EPP (to enable flow-control metrics, e.g. saturation), and a real Prometheus instance.All e2e test scenarios (general integration, saturation gate, budget gate, redis gate) now run against this stack.

All e2e test scenarios (general integration, saturation gate, budget gate, redis gate) now run against this stack.

Key changes:
- Remove `igw-mock/` and `prom-mock/` and their k8s manifests
- Add k8s manifests for sim, envoy, EPP (with and without flow control), and Prometheus (with `inference_pool` relabeling)
- Deploy async-processor instances via the helm chart (4 releases: integration, saturation, budget, redis-gate) instead of a single raw YAML manifest —
this also validates the chart in CI
- Rewrite test helpers to drive saturation/budget through the sim's admin API and wait for metrics to propagate through EPP → Prometheus
- Add Envoy fault injection helpers for retry testing
- Budget cascade test: verify fallback from EPP flow-control metrics to vLLM saturation metrics, then recovery when EPP is restarted with flow control
enabled

### Running the e2e suite

```bash
make test-e2e
```

Requires: docker (or podman), kind, helm, kubectl. The suite creates a disposable Kind cluster, builds and loads all images, deploys the full stack, and
  tears down on completion. Set `E2E_SKIP_CLEANUP=true` to keep the cluster for debugging.

## Why is this change needed?

The mock-based e2e tests do not catch integration issues between the async processor and real upstream components (EPP metric computation, Prometheus
  scraping, PromQL evaluation). By testing against the actual inference pipeline, we gain confidence that the gating logic works end-to-end — particularly
  the cascade from EPP flow-control metrics to vLLM saturation metrics, which involves Prometheus scrape timing, metric relabeling, and PromQL vector
  matching that mocks cannot reproduce.

## How was this tested?

## Checklist

## Related Issues

Follow up to #106, #145, #163


evacchi:
I am going to wait until GAIE is fully merged into llm-d-inference-scheduler

evacchi:
maybe for the simulator we can just use `ghcr.io/llm-d/llm-d-inference-sim:v0.0.0-test`

shimib (dismissed):

evacchi:
ok the top post was updated. This has become sort of large but in short it's 2 PRs in one, and I can split it if it's easier to review. Check the top post.

lioraron:
Good work!

Several issues to address:

### 1. `defer resp.Body.Close()` inside a for loop (`e2e_suite_test.go`, `fetchGAIECRDs`)

```go
for i, name := range gaieInferencePoolCRDs {
    resp, err := http.Get(baseURL + "/" + name)
    // ...
    defer resp.Body.Close() // ← deferred to function return, not loop iteration
```

All response bodies accumulate until `fetchGAIECRDs` returns. Use an immediately-invoked closure or close inline after `ReadAll`.

### 2. Floating-point exact-zero comparison in budget gate tests

In `e2e_budget_gate_test.go`, both "blocks messages" and "resumes processing" tests use:

```go
return v >= 0 && v <= 0  // ← checking float == 0.0
```

Floating-point equality is fragile for a multi-hop metric propagation chain (sim → EPP → Prometheus → query). Use a small epsilon: `v >= 0 && v < 0.05` or similar. The baseline in the budget values is already `0.05`, so the gate closes when `D ≤ baseline` — you don't need exact zero to validate the gate is closed.

### 3. Massive duplication between `epp-fc.yaml` and `epp-no-fc.yaml`

These two 178/190-line manifests share ~90% of content (ServiceAccount, Role, RoleBinding, ClusterRole, ClusterRoleBinding, Deployment, Service). Only the ConfigMap and one env var differ. Consider:
- A single manifest with a `${FLOW_CONTROL_ENABLED}` template variable
- Or factor the shared RBAC/Service resources into a base manifest applied once, with the ConfigMap+Deployment as the only toggled piece

The current duplication means any change to RBAC, ports, probe config, or resource limits needs to be applied in both files.

### 4. Redundant `redeployEPPWithFlowControl()` calls

Both `Saturation Metric Dispatch Gate E2E` and `Budget Metric Dispatch Gate E2E` suites call `redeployEPPWithFlowControl()` in their `BeforeAll`. Since these are `ginkgo.Ordered`, the second suite redeploys EPP (rollout restart + wait for all 4 processor rollouts) for no reason — it was already deployed with flow control by the first suite's `BeforeAll`. This adds ~2 minutes of unnecessary wait time. Consider moving the redeploy to a shared `BeforeAll` or making it idempotent (check if FC is already enabled before redeploying).

### 5. Hardcoded namespace in all YAML manifests

Every YAML manifest hardcodes `namespace: e2e-integration`, while the test suite defines `nsName = "e2e-integration"` as a constant and `kubectlApplyFile` already supports template substitutions. Using `${NAMESPACE}` in the manifests (like you already do for images) would be more consistent and allow namespace customization.

### 6. Chart changes should be a separate PR

The Helm chart enhancements (gate params JSON rendering, Redis secret auto-creation, PodMonitor template, new values for `concurrency`, `prometheusURL`, `prometheusCacheTTL`, `pollIntervalMs`, `batchSize`, `gateType`, `gateParams`) are independently valuable and reviewable. Bundling them with the e2e test rewrite makes both harder to review. As you noted in the issue comments, this is "2 PRs in one" — splitting would help.

### Minor

- `splitImage` in `e2e_suite_test.go` uses `strings.LastIndex(image, ":")` which would mis-split digest references (`repo@sha256:abc` → repo=`repo@sha256`, tag=`abc`). Not a problem today since all test images use tags, but worth a comment noting this limitation.
- `os.Setenv("KUBECONFIG", kindKubeconfig)` in `setupK8sClient` is a global side effect with `//nolint:errcheck`. Consider using `t.Setenv` (available via ginkgo) or at least checking the error.

evacchi:
will address comments + spin off the helm charts+doc-related changes to their own PR

evacchi:
doc changes now in #163 -- redrafting this

lioraron (commented):
test

test/e2e/utils_test.go
```diff
@@ -67,102 +77,169 @@ func popResult(ctx context.Context, rdb *redis.Client, queue string) *api.Result
...
+	body := []byte(`{"model":"test-model","prompt":"probe"}`)
+	req, err := http.NewRequest(http.MethodPost, envoyURL+"/v1/completions", bytes.NewReader(body))
+	if err != nil {
+		return
+	}
 	req.Header.Set("Content-Type", "application/json")
-	resp, err := adminClient.Do(req)
-	gomega.ExpectWithOffset(1, err).NotTo(gomega.HaveOccurred())
+	resp, err := httpClient.Do(req)
+	if err != nil {
```
lioraron, line 118: Both error paths silently return with no logging. When Envoy is unreachable, callers like `waitForBudget` and `waitForSaturation` time out with a misleading "waiting for budget to satisfy condition" message — no indication that probes are actually failing.

Consider adding debug-level logging:
```go
if err != nil {
    ginkgo.GinkgoLogr.V(1).Info("probe request failed", "error", err)
    return
}
```
Same for the connection error path above.

test/e2e/utils_test.go
```diff
@@ -67,102 +77,169 @@ func popResult(ctx context.Context, rdb *redis.Client, queue string) *api.Result
...
+// queryProm runs a PromQL instant query and returns the first result's scalar
+// value. Returns NaN if the query fails or returns no results; callers must
+// check with math.IsNaN. Do NOT use a numeric sentinel like -1: budget PromQL
+// (1 - queue_size/capacity) legitimately returns negative values when queue_size
+// exceeds capacity.
+func queryProm(promURL, query string) float64 {
+	resp, err := httpClient.Get(promURL + "/api/v1/query?query=" + url.QueryEscape(query))
+	if err != nil {
+		return math.NaN()
+	}
```
lioraron, line 135: HTTP status code is never checked. If Prometheus returns 500/503 (e.g., during startup or OOM), `json.Decode` fails silently and returns `NaN` — all dependent `Eventually` loops time out with no diagnostic output.

Suggestion: check status and log non-200 responses before decoding:
```go
if resp.StatusCode != http.StatusOK {
    ginkgo.GinkgoLogr.V(1).Info("prom query returned non-200",
        "status", resp.StatusCode, "query", query)
    return math.NaN()
}
```

test/e2e/e2e_suite_test.go
```diff
@@ -186,109 +220,342 @@ func setupK8sClient() {
...
 		return rdb.Ping(context.Background()).Err()
 	}, 30*time.Second, 1*time.Second).Should(gomega.Succeed())
 
-	ginkgo.By("Waiting for prom-mock to be ready on localhost:" + promMockPort)
+	ginkgo.By("Waiting for Prometheus to be ready")
 	gomega.Eventually(func() error {
-		resp, err := http.Get(promMockURL + "/admin/saturation")
+		resp, err := http.Get(promURL + "/-/ready")
 		if err != nil {
 			return err
```
lioraron: The readiness checks for Prometheus, sim, and Envoy only check for connection errors — they don't verify HTTP status codes. If Prometheus returns 503 on `/-/ready` (its explicit "not ready" signal), this check passes and the suite proceeds before Prometheus is actually ready, potentially causing flaky metric queries.

Suggestion:
```go
if resp.StatusCode != http.StatusOK {
    return fmt.Errorf("prometheus not ready: %d", resp.StatusCode)
}
```

evacchi:
realizing there were some e2e tests missing in the port, working on it.
edit: done

lioraron (approved):
Thanks!

shimib (approved):

PR #139, opened 2026-04-29
feat: Add per-attribute quota management for pubsub/redis

shimib:
Implemented a flexible 'AttributeGate' interface to support granular, per-user/team quota control.

- Added 'RedisQuotaGate' supporting sliding window rate limiting and TTL-backed concurrency management via atomic Lua scripts.
- Integrated 'AttributeGate' into GCP Pub/Sub flow, nacking messages when quota is exceeded to leverage native backpressure.
- Integrated 'AttributeGate' into Redis Sorted Set flow, safely maintaining a release hook across the processing lifecycle.
- Created robust unit tests and an E2E test to verify strict concurrency logic with a mock inference gateway.
- Updated GateFactory to allow initialization from JSON configurations natively.

## How was this tested?

## Checklist


pkg/redis/quota_gate.go
```diff
@@ -0,0 +1,139 @@
...
+	res, err := g.rdb.Eval(ctx, script, []string{key}, g.limit, ttl).Result()
+	if err != nil {
+		return false, nil, err
+	}
+
+	if res.(int64) == 0 {
+		return false, nil, nil
+	}
+
+	release := func() {
```
crissepulveda-0, line 98: I could be misunderstanding exactly how context.Background would work in this situation, but from what I can tell: DECR will happen here even if the request is canceled. I’m concerned that if the processor crashes or a network error happens during this call, the "release" might never reach Redis.

If that happens, does this user's quota stay "full" forever? To prevent this, should we update the Lua script on Line 72 to add an automatic expiration timer to the key?

For example, if we set it to expire after window * 2, the quota would automatically reset itself even if a release() call is missed.

pkg/pubsub/pubsubimpl.go
```diff
@@ -278,43 +278,69 @@ func (r *PubSubMQFlow) requestWorker(ctx context.Context, pubSubClient *pubsub.C
...
-				msg.Ack()
+				return
 			}
-		})
-		cancel()
-		// TODO
-		if err != nil {
-			logger.V(logutil.DEFAULT).Error(err, "Fail to receive messages from request subscription")
+			if !allowed {
+				msg.Nack()
```
crissepulveda-0: In GCP Pub/Sub, a Nack tells the system to redeliver the message almost immediately. If a user is completely over their limit and has a bunch of messages lined up, won't this cause a fast retry loop that burns through network bandwidth? Ref: https://docs.cloud.google.com/pubsub/docs/pull-troubleshooting#subscriber_negatively_acknowledged_the_messages

Maybe we should use msg.ModAckDeadline(10) instead to push the retry back by 10 seconds? That way, we give their active requests a real chance to finish before trying again."

crissepulveda-0:
For some reason I am getting an error when I try to leave comments on specific lines for the review. Ran into this yesterday and thought it would be fixed by now. Leaving the rest of my comments below:

---

pkg/redis/sortedset_impl.go
Line: 291

To keep the state totally clean, should we move r.activeReleases.Store(...) inside the actual case msgChannel <- ir: block? That way we avoid a situation where the channel block times out or gets cancelled right after, and we run into a weird state where we call the release manually but the map might still hold onto it. We should only track it once we're 100% sure the worker has the message

---

pkg/redis/quota_gate.go
Line: 58
The key format %s%s:%s (e.g., quota:userid:123) might lead to collisions if two different gates use the same prefix but different attributes.Can we change the key structure to something more explicit like %sattr:%s:val:%s? This makes the Redis database much easier to debug and avoids accidental overlaps.

---

README.md
Line: 155

It might be helpful to explicitly mention that the gates parameter is a JSON-string-inside-JSON, as the escaping (\") can be a bit tricky for users to get right on the first try. I remember running into similar issues with customers not understanding documentation I wrote for GKE Autopilot Allowlisting.

crissepulveda-0 (approved):

PR #142, opened 2026-04-30
refactor: Introduce Request interface and InternalRouting struct for per message queue routing fields

vishbhat:
## What does this PR do?

Introduces a `Request` interface and `InternalRouting` struct to cleanly separate the caller-facing request contract from system-internal routing concerns.
**Key changes:**
- Defines a `Request` interface with concrete types: `RequestMessage`, `RedisRequest`, `PubSubRequest`
- Introduces `InternalRouting` struct to hold resolved routing fields (queue names, retry count, correlation ID) that were previously smuggled through metadata maps
- Wraps both in an `InternalRequest` envelope with tagged-envelope JSON serialization (`request_kind` discriminator)
- Changes `RequestChannel` from `chan RequestMessage` to `chan *InternalRequest`
- Changes deadline/created fields from `string` to `int64`, removing scattered `strconv.ParseInt` calls
- Removes legacy flat JSON deserialization (no backward compat needed — repo is pre-release)
- Removes metadata-based routing hacks (`QUEUE_NAME_KEY`, `SORTEDSET_QUEUE_NAME_KEY`, `PUBSUB_ID` constants)
- Updates `Producer` interface to accept `Request` instead of `RequestMessage`
- Updates README documentation to reflect the new request format

## Why is this change needed?

The previous design conflated user-supplied data with infrastructure routing by storing queue names, pubsub IDs, and retry counts in the request's `Metadata` map. This made the code fragile (magic string keys), hard to reason about (unclear what's caller data vs system data), and prevented per-message queue routing without polluting the public API. Deadlines stored as strings required repetitive parsing at every consumption point.

This refactor establishes a clear boundary: callers interact with the `Request` interface (id, deadline, payload, metadata), while producers and workers operate on `InternalRouting` for queue resolution, retries, and correlation — making the system easier to extend and less error-prone.

## How was this tested?

image

## Checklist

## Related Issues

Related to https://github.com/llm-d-incubation/llm-d-async/issues/121


pkg/redis/sortedset_impl.go
```diff
@@ -318,12 +323,16 @@ func (r *RedisSortedSetFlow) flushRetryBatch(ctx context.Context, batch []api.Re
 
 	entries := make([]retryEntry, 0, len(batch))
 	for _, msg := range batch {
-		queueName := msg.Metadata[SORTEDSET_QUEUE_NAME_KEY]
+		queueName := msg.RequestQueueName
```
lioraron: **Bug — nil-pointer dereference before nil check**

`msg.RequestQueueName` dereferences the embedded `*InternalRequest` pointer (chain: `RetryMessage` → `EmbelishedRequestMessage` → `*InternalRequest` → `InternalRouting.RequestQueueName`). If `InternalRequest` is nil, this panics before the guard on line 331 catches it.

Move the nil check before the field access:
```go
for _, msg := range batch {
	if msg.InternalRequest == nil {
		logger.V(logutil.DEFAULT).Error(nil, "Retry message missing InternalRequest")
		continue
	}
	queueName := msg.RequestQueueName
	if queueName == "" {
		queueName = *ssRequestQueueName
	}
	// ...
}
```

lioraron (commented):
Review from code inspection.

pkg/redis/sortedset_impl.go
vishbhat: @lioraron Thats a good catch, will fix it. Thanks!

api/api.go
```diff
@@ -61,19 +61,60 @@ type RequestMergePolicy interface {
 	MergeRequestChannels(channels []RequestChannel) EmbelishedRequestChannel
 }
 
-// TODO: Consider per-message metadata map[string]string
-// add endpoint to message level.
+// Request is the public interface for submitting requests to the async queue.
+// It exposes only the caller-visible fields. Concrete types like RequestMessage,
+// RedisRequest, and PubSubRequest satisfy this interface.
+type Request interface {
+	ReqId() string
```
lioraron: Per golang conventions `ReqID() string`.

api/api.go
```diff
@@ -61,19 +61,60 @@ type RequestMergePolicy interface {
...
+// Request interface accessors use the Req prefix (e.g. ReqPayload) to avoid
+// colliding with the struct's exported field names used for JSON serialization.
 type RequestMessage struct {
-	Id              string            `json:"id"`
-	CreatedUnixSec  string            `json:"created"`               // Unix seconds
-	RetryCount      int               `json:"retry_count,omitempty"` // TODO: Consider
-	DeadlineUnixSec string            `json:"deadline"`              // TODO: check about using int64, change name to timeout
-	Payload         map[string]any    `json:"payload"`
-	Metadata        map[string]string `json:"metadata,omitempty"`
+	Id       string            `json:"id"`
```
lioraron: Per golang conventions `ID string`.

api/api.go
```diff
@@ -84,21 +125,27 @@ type EmbelishedRequestChannel struct {
 	Channel chan EmbelishedRequestMessage
 }
 
+// EmbelishedRequestMessage decorates an InternalRequest with HTTP dispatch context.
```
lioraron: **Pre-existing, not introduced by this PR, but important:** `RequestChannel` now carries `chan *InternalRequest`, which leaks an internal/infrastructure type into the `api` package — the same package that defines the caller-facing `Request` interface and `RequestMessage`.

`RequestChannel`, `EmbelishedRequestMessage`, `RetryMessage`, etc. are internal plumbing (used by queue implementations and the merge policy, not by external callers). Consider splitting the public contract (`Request`, `RequestMessage`, `Producer`) from the internal pipeline types into a separate package (e.g. `internal/pipeline`).

api/api.go
```diff
@@ -84,21 +125,27 @@ type EmbelishedRequestChannel struct {
...
 type RetryMessage struct {
 	EmbelishedRequestMessage
 	BackoffDurationSeconds float64
 }
 
-// optional field of httpstatus, golang error?
+// ResultMessage is the async inference result returned to callers. Id and Payload are
+// JSON fields; Routing and Metadata are infrastructure pass-through (json:"-").
 type ResultMessage struct {
 	Id       string            `json:"id"`
```
lioraron: ID

api/internal_api.go
```diff
@@ -0,0 +1,125 @@
+package api
```
lioraron, line 1: This file should be in a separate place, as it's internal, and not related to the user facing producer api.

api/api.go
vishbhat: Will change it. Just FYI: it was pre-existing, not introduced in this PR.

api/api.go
vishbhat: For now I have split the split the files based on user facing and internal API. Will create separate package in a follow up PR. Tracking it via: https://github.com/llm-d-incubation/llm-d-async/issues/144

api/internal_api.go
vishbhat: Will be handled in https://github.com/llm-d-incubation/llm-d-async/issues/144

shimib (approved):

PR #145, opened 2026-05-04
feat: update dispatch budget gate 

evacchi:
note this PR is stacked on #140 so you can ignore the docs/dispatch-budget.md change

## What does this PR do?

implements the changes detailed in #140; the dispatch budget now is computed using a metric from the EPP (queue depth or saturation. In this impl, queue depth.) and if missing (e.g. flow control is disabled) it uses a metric of _saturation_ equivalent to that of the EPP, but computed via Prometheus

## Why is this change needed?

- we implement the new formula described in #140 
- we adopt a cascade strategy in case flow control is disabled (default config) but the users deploy the batch gateway 

## How was this tested?

includes mock tests that will be replaced in #115 

## Checklist

## Related Issues

related + blocked: #140

### DIFF
--- a/pkg/async/inference/flowcontrol/cascade_metric_source.go
+++ b/pkg/async/inference/flowcontrol/cascade_metric_source.go
@@ -29,32 +29,37 @@ var _ MetricSource = (*CascadeMetricSource)(nil)
 
 // CascadeMetricSource tries each source in order and returns the first successful result.
 // If a source returns an error or no samples, the next source is tried. Log messages are
-// emitted only on transitions (entering/leaving fallback) to avoid noise during sustained outages.
+// emitted only on transitions (entering/leaving fallback, or switching fallback source)
+// to avoid noise during sustained outages.
 type CascadeMetricSource struct {
-	sources       []MetricSource
-	usingFallback atomic.Bool
+	sources     []MetricSource
+	activeIndex atomic.Int32 // 0 = primary, >0 = fallback index
 }
 
 // NewCascadeMetricSource creates a CascadeMetricSource from the given sources, tried in order.
-// At least two sources are required.
+// At least two sources are required; panics otherwise.
 func NewCascadeMetricSource(sources ...MetricSource) *CascadeMetricSource {
+	if len(sources) < 2 {
+		panic("CascadeMetricSource requires at least two sources")
+	}
 	return &CascadeMetricSource{sources: sources}
 }
 
-// Query implements MetricSource. It tries each source in order, logging once when
-// entering fallback and once when recovering to the primary.
+// Query implements MetricSource. It tries each source in order, logging on transitions:
+// entering fallback, switching fallback source, or recovering to the primary.
 func (c *CascadeMetricSource) Query(ctx context.Context) ([]Sample, error) {
 	logger := log.FromContext(ctx)
 	for i, s := range c.sources {
 		samples, err := s.Query(ctx)
 		if err == nil && len(samples) > 0 {
-			if i > 0 {
-				if c.usingFallback.CompareAndSwap(false, true) {
-					logger.V(logutil.DEFAULT).Info("primary metric source unavailable, using fallback",
-						"fallbackIndex", i)
+			prev := c.activeIndex.Swap(int32(i))
+			if int32(i) != prev {
+				if i == 0 {
+					logger.V(logutil.DEFAULT).Info("primary metric source recovered")
+				} else {
+					logger.V(logutil.DEFAULT).Info("using fallback metric source",
+						"fallbackIndex", i, "previousIndex", prev)
 				}
-			} else if c.usingFallback.CompareAndSwap(true, false) {
-				logger.V(logutil.DEFAULT).Info("primary metric source recovered")
 			}
 			return samples, nil
 		}
--- a/pkg/async/inference/flowcontrol/cascade_metric_source.go
+++ b/pkg/async/inference/flowcontrol/cascade_metric_source.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2026 The llm-d Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flowcontrol
+
+import (
+	"context"
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	logutil "sigs.k8s.io/gateway-api-inference-extension/pkg/epp/util/logging"
+)
+
+var _ MetricSource = (*CascadeMetricSource)(nil)
+
+// CascadeMetricSource tries each source in order and returns the first successful result.
+// If a source returns an error or no samples, the next source is tried and a log message
+// is emitted so operators can detect when the cascade is active.
+type CascadeMetricSource struct {
+	sources []MetricSource
+}
+
+// NewCascadeMetricSource creates a CascadeMetricSource from the given sources, tried in order.
+// At least two sources are required.
+func NewCascadeMetricSource(sources ...MetricSource) *CascadeMetricSource {
+	return &CascadeMetricSource{sources: sources}
+}
+
+// Query implements MetricSource. It tries each source in order, logging a warning when
+// falling back so operators can detect degraded metric availability.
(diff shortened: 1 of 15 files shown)


pkg/async/inference/flowcontrol/cascade_metric_source.go
```diff
@@ -0,0 +1,61 @@
...
+			}
+			return samples, nil
+		}
+		if err != nil {
+			logger.V(logutil.DEFAULT).Info("metric source unavailable, trying next",
+				"sourceIndex", i, "error", err)
+		}
+	}
+	return nil, fmt.Errorf("all metric sources unavailable")
+}
```
lioraron, line 68: **Cascade logs on every `Query()` call when primary is down**

When the primary source is unavailable, this logs two messages per call: "metric source unavailable, trying next" (line 57) and "primary metric source unavailable, using fallback" (line 51). The `CachedMetricSource` caches errors too, so it returns the cached error quickly — but the cascade still processes it on every `Budget()` call from the gate.

With the e2e poll interval of 500ms, that's ~4 log lines/second continuously while the primary is down. In production with longer outages this could be noisy.

`MetricDispatchGate` already solved this same problem with the `usingFallback` atomic.Bool pattern (nice addition in this PR). Consider applying the same pattern here — log once on transition into fallback, once on recovery, and stay quiet in between.

pkg/async/inference/flowcontrol/cascade_metric_source.go
```diff
@@ -0,0 +1,63 @@
...
+// CascadeMetricSource tries each source in order and returns the first successful result.
+// If a source returns an error or no samples, the next source is tried. Log messages are
+// emitted only on transitions (entering/leaving fallback) to avoid noise during sustained outages.
+type CascadeMetricSource struct {
+	sources       []MetricSource
+	usingFallback atomic.Bool
+}
+
+// NewCascadeMetricSource creates a CascadeMetricSource from the given sources, tried in order.
+// At least two sources are required.
```
acardace: nit: I don't see anything enforcing this, so maybe the comment should be removed?

pkg/async/inference/flowcontrol/cascade_metric_source.go
```diff
@@ -0,0 +1,63 @@
...
+// Query implements MetricSource. It tries each source in order, logging once when
+// entering fallback and once when recovering to the primary.
+func (c *CascadeMetricSource) Query(ctx context.Context) ([]Sample, error) {
+	logger := log.FromContext(ctx)
+	for i, s := range c.sources {
+		samples, err := s.Query(ctx)
+		if err == nil && len(samples) > 0 {
+			if i > 0 {
+				if c.usingFallback.CompareAndSwap(false, true) {
+					logger.V(logutil.DEFAULT).Info("primary metric source unavailable, using fallback",
```
acardace: You don't log again if you change the fallback source, i.e. you go from source 1 to source 2. I think probably a log would be useful in that case as well? Maybe you can store the fallback index instead of using a simple bool. https://pkg.go.dev/sync/atomic#Int32.CompareAndSwap

pkg/async/inference/flowcontrol/gate_factory.go
```diff
@@ -119,21 +123,42 @@ func (f *GateFactory) CreateGate(gateType string, params map[string]string) (asy
...
+			return nil, fmt.Errorf("inference pool name is required for prometheus-budget gate")
+		}
+		maxConcurrency, err := parseFloat("max_concurrency", params["max_concurrency"], 100.0)
+		if err != nil {
+			return nil, err
+		}
+		if maxConcurrency <= 0 {
+			return nil, fmt.Errorf("max_concurrency must be positive, got %g", maxConcurrency)
+		}
+		baseline, err := parseFloat("baseline", params["baseline"], 0.05)
```
acardace, line 137: nit: test that baseline sits in a valid range? [0, 100) ?

pkg/async/inference/flowcontrol/gate_factory_test.go
```diff
@@ -150,64 +150,56 @@ func TestGateFactory_RedisGateDifferentAddresses(t *testing.T) {
...
-	assert.Contains(t, err.Error(), "max_sys must be positive")
+	assert.NoError(t, err)
+	assert.NotNil(t, gate)
 }
 
 func TestGateFactory_BudgetGateWithInvalidBaseline(t *testing.T) {
 	factory := NewGateFactory("http://localhost:9090")
 	gate, err := factory.CreateGate("prometheus-budget", map[string]string{
-		"max_sys":  "50",
+		"pool":     "my-pool",
```
acardace, line 202: nit: if you introduce a valid range for baseline add a test too.

pkg/async/inference/flowcontrol/metric_dispatch_gate.go
```diff
@@ -49,45 +52,58 @@ func NewMetricDispatchGate(source MetricSource, threshold float64, fallback floa
...
 // On error or missing data the gate returns the configured fallback budget.
-// The output is always clamped to [0.0, 1.0].
 func (g *MetricDispatchGate) Budget(ctx context.Context) float64 {
 	logger := log.FromContext(ctx)
 
 	samples, err := g.source.Query(ctx)
 	if err != nil {
-		logger.V(logutil.DEFAULT).Info("MetricSource error, using fallback value", "fallback", g.fallback, "error", err)
+		if g.usingFallback.CompareAndSwap(false, true) {
+			logger.V(logutil.DEFAULT).Info("MetricSource error, using fallback value", "fallback", g.fallback, "error", err)
```
acardace: I think this should be an Error level log, using fallback here means nothing is getting dispatched.

acardace: Not sure rate-limiting so hard is the right choice here, I prefer to be spammed is something is really stuck and needs intervention.

pkg/async/inference/flowcontrol/metric_dispatch_gate.go
```diff
@@ -49,45 +52,58 @@ func NewMetricDispatchGate(source MetricSource, threshold float64, fallback floa
...
+		}
 		return g.fallback
 	}
 
 	if len(samples) == 0 {
-		logger.V(logutil.DEFAULT).Info("No metric samples found, using fallback value", "fallback", g.fallback)
+		if g.usingFallback.CompareAndSwap(false, true) {
+			logger.V(logutil.DEFAULT).Info("No metric samples found, using fallback value", "fallback", g.fallback)
```
acardace: Maybe this is not an Error but definitely a warning?

pkg/async/inference/flowcontrol/metric_dispatch_gate.go
```diff
@@ -49,45 +52,58 @@ func NewMetricDispatchGate(source MetricSource, threshold float64, fallback floa
...
+		}
 		return g.fallback
 	}
 
 	value := samples[0].Value
 	if math.IsNaN(value) || math.IsInf(value, 0) {
-		logger.V(logutil.DEFAULT).Info("Invalid metric value, using fallback value", "fallback", g.fallback, "value", value)
+		if g.usingFallback.CompareAndSwap(false, true) {
+			logger.V(logutil.DEFAULT).Info("Invalid metric value, using fallback value", "fallback", g.fallback, "value", value)
```
acardace: Same here, Warning/Error?

test/e2e/e2e_budget_gate_test.go
```diff
@@ -22,8 +22,12 @@ var _ = ginkgo.Describe("Budget Metric Dispatch Gate E2E", func() {
...
 		rdb.Del(ctx, budgetResultQueue)  //nolint:errcheck
 		resetMock(adminURL)
-		// Start with full dispatch budget (no load)
-		setPromMockBudget(promMockURL, "1.0")
+		// Reset all prom-mock metric sources (primary budget, vLLM budget,
+		// primaryDisabled flag) so cascade state doesn't leak between tests.
+		resetPromMock(promMockURL)
+		// Wait for the processor's metric cache (5s TTL) to expire and
+		// pick up the reset (gate-closed) values before each test runs.
+		time.Sleep(6 * time.Second)
```
acardace: Caching is disabled in tests https://github.com/llm-d-incubation/llm-d-async/blob/main/test/e2e/yaml/async-processor-budget.yaml#L53, no need to sleep 6 seconds.

pkg/async/inference/flowcontrol/cascade_metric_source.go
evacchi: now it's enforced in the constructor

pkg/async/inference/flowcontrol/cascade_metric_source.go
evacchi: done

pkg/async/inference/flowcontrol/metric_dispatch_gate.go
evacchi: done

evacchi:
merge https://github.com/llm-d-incubation/llm-d-async/pull/140 before this

shimib (approved):

PR #161, opened 2026-05-06
fix: add WithRedisClient option to allow injecting custom Redis Client in Producer

### LINKED ISSUE #157: [Bug]: NewRedisSortedSetProducer creates Redis client internally, preventing tracing instrumentation
## What happened?

`NewRedisSortedSetProducer` creates its own `redis.Client` internally (redis_sortedset_producer.go:78-82), so callers cannot inject an instrumented client. This makes it impossible to add distributed tracing (e.g., `redisotel.InstrumentTracing`) or any other Redis hook-based instrumentation to the producer.

## Expected behavior

Callers should be able to provide their own `redis.Client` so they can instrument it with OpenTelemetry tracing, custom metrics, or other hooks before passing it in.

Suggested approach: add a functional option `WithRedisClient(*redis.Client)` that uses the provided client instead of creating one internally. When not set, the current behavior is preserved (backward compatible).

### Example usage

```go
// Create and instrument a Redis client with OpenTelemetry
client := redis.NewClient(&redis.Options{
    Addr:     "localhost:6379",
    Username: "user",
    Password: "pass",
})
redisotel.InstrumentTracing(client)
redisotel.InstrumentMetrics(client)

// Inject the instrumented client into the producer
producer, err := producer.NewRedisSortedSetProducer(
    producer.RedisSortedSetConfig{
        TenantID: "tenant-abc",
    },
    producer.WithRedisClient(client),
)
```

### Suggested implementation

```go
type Option func(*RedisSortedSetProducer)

func WithRedisClient(client *redis.Client) Option {
    return func(p *RedisSortedSetProducer) {
        p.client = client
    }
}

func NewRedisSortedSetProducer(config RedisSortedSetConfig, opts ...Option) (*RedisSortedSetProducer, error) {
    // ... validation ...

    p := &RedisSortedSetProducer{
        requestQueueName: config.RequestQueueName,
        resultQueueName:  namespacedResultQueue,
    }

    for _, opt := range opts {
        opt(p)
    }

    // Only create a client if one was not injected
    if p.client == nil {
        if config.RedisAddr == "" {
            return nil, errors.New("RedisAddr is required when no RedisClient is provided")
        }
        p.client = redis.NewClient(&redis.Options{
            Addr:     config.RedisAddr,
            Username: config.RedisUser,
            Password: config.RedisPassword,
        })
    }

    // Test connection
    ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
    defer cancel()
    if err := p.client.Ping(ctx).Err(); err != nil {
        return nil, fmt.Errorf("failed to connect to Redis: %w", err)
    }

    return p, nil
}
```

## Steps to Reproduce

1. Try to add OpenTelemetry tracing to Redis operations in the producer
2. There is no way to call `redisotel.InstrumentTracing(client)` because the client is created inside `NewRedisSortedSetProducer` and not exposed

vishbhat:
## What does this PR do?

- Adds a `WithRedisClient(*redis.Client)` functional option to `NewRedisSortedSetProducer`, allowing callers to inject a pre-configured Redis client instead of having one created internally
- When a client is provided via `WithRedisClient`, `RedisURL` is no longer required in the config
- When no client is injected, existing behaviour is fully preserved (backward compatible)

## Why is this change needed?

`NewRedisSortedSetProducer` creates its Redis client internally, making it impossible to add hooks like OpenTelemetry tracing (`redisotel.InstrumentTracing`) or custom metrics before the client is used. This change enables distributed tracing and observability instrumentation by letting callers bring their own instrumented client (#157).

## How was this tested?

## Checklist

## Related Issues

Fixes #157

### DIFF
--- a/producer/redis_sortedset_producer.go
+++ b/producer/redis_sortedset_producer.go
@@ -22,6 +22,18 @@ type RedisSortedSetProducer struct {
 	resultQueueName  string
 }
 
+// ProducerOption is a functional option for NewRedisSortedSetProducer.
+type ProducerOption func(*RedisSortedSetProducer)
+
+// WithRedisClient injects a pre-configured *redis.Client, allowing callers to
+// instrument it (e.g. with OpenTelemetry tracing/metrics hooks) before use.
+// When provided, RedisURL in the config is not required.
+func WithRedisClient(client *redis.Client) ProducerOption {
+	return func(p *RedisSortedSetProducer) {
+		p.client = client
+	}
+}
+
 // RedisSortedSetConfig contains configuration for the Redis sorted set producer.
 type RedisSortedSetConfig struct {
 	// RedisURL is a Redis URL (e.g. "redis://user:pass@host:port/db" or "rediss://..." for TLS).
@@ -47,11 +59,8 @@ type RedisSortedSetConfig struct {
 
 // NewRedisSortedSetProducer creates a new producer using Redis sorted set.
 // Multi-tenant safe: TenantID ensures result queue isolation between tenants.
-func NewRedisSortedSetProducer(config RedisSortedSetConfig) (*RedisSortedSetProducer, error) {
-	if config.RedisURL == "" {
-		return nil, errors.New("RedisURL is required")
-	}
-
+// Use WithRedisClient to inject a pre-configured client (e.g. with tracing hooks).
+func NewRedisSortedSetProducer(config RedisSortedSetConfig, opts ...ProducerOption) (*RedisSortedSetProducer, error) {
 	if config.TenantID == "" {
 		return nil, errors.New("TenantID is required for multi-tenant isolation")
 	}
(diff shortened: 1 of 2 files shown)


producer/redis_sortedset_producer.go
```diff
@@ -22,6 +22,18 @@ type RedisSortedSetProducer struct {
...
 
+// ProducerOption is a functional option for NewRedisSortedSetProducer.
+type ProducerOption func(*RedisSortedSetProducer)
+
+// WithRedisClient injects a pre-configured *redis.Client, allowing callers to
+// instrument it (e.g. with OpenTelemetry tracing/metrics hooks) before use.
+// When provided, RedisURL in the config is not required.
+func WithRedisClient(client *redis.Client) ProducerOption {
+	return func(p *RedisSortedSetProducer) {
+		p.client = client
```
yizhaodev, line 34: if the client is passed by external, you should not close it because it might be a shared client

vishbhat: Thats a valid point. Ill fix it.

shimib (approved):

lioraron:
Good work!

### 1. First commit missing DCO sign-off

Commit `3d7af7b` ("fix: add WithRedisClient option...") has no `Signed-off-by` trailer. The second commit (`e944fb9`) has it. The PR checklist says "Commits are signed off" but this will likely fail the DCO check. Squash or amend the first commit to add the sign-off.

### 2. Test accesses private field `p.client`

In `TestWithRedisClient`, the "injected client is used" subtest directly accesses `p.client` to call `Ping`:

```go
err = p.client.Ping(ctx).Err()
```

This reaches into the struct's internals. Either make this a public method (unlikely to be wanted), or validate through the public API — e.g. call `p.SubmitRequest(...)` with a valid request and verify it succeeds. That proves the injected client is wired up without depending on the struct layout.

### 3. No test for `Close()` ownership semantics

The `managedClient` flag is the key behavioral addition (addressing yizhaodev's feedback), but there's no test for it. Add two cases:

- **Managed client**: create producer with `RedisURL`, call `Close()`, verify client is closed (e.g. subsequent `Ping` returns an error)
- **Injected client**: create producer with `WithRedisClient(client)`, call `Close()`, verify the injected client is still usable (Ping succeeds)

This is the whole point of the `managedClient` distinction — it should have a test.

### 4. `WithRedisClient(nil)` silently falls through to URL-based creation

If someone writes `WithRedisClient(nil)`, the option sets `p.client = nil`, then the constructor falls through to the `if p.client == nil` branch and tries to create from `RedisURL`. This is arguably fine, but could mask a caller bug where they intended to inject a client but passed nil by accident. Consider either:
- Documenting this behavior in the `WithRedisClient` doc comment, or
- Returning an error if the injected client is nil (`errors.New("WithRedisClient: client must not be nil")`)

### Minor

- The `RedisURL` field doc comment still says `// Required.` — should be updated to say "Required unless a client is injected via WithRedisClient".

PR #192, opened 2026-05-14
feat: simplified config in Helm

evacchi:
## What does this PR do?

Adds support for full multi-queue JSON configuration in the Helm chart via a new queuesConfig values field. Previously the chart only supported a single queue with a single dispatch gate through individual CLI flags.

### Go changes:
- Adds --redis.ss.queues-config and --redis.queues-config inline CLI flags that accept the full queues JSON directly (priority: inline > file > single-queue fallback)
- Adds a StringMap type that coerces non-string JSON values (e.g. threshold: 0.7) by coercing them to strings during unmarshaling, so Helm's toJson works without a custom helper
   - Note: this is a stopgap measure because GateParams is `map[string]string`, we should follow-up with a proper refactor to take at least `map[string]any`.

### Helm changes:
- Adds queuesConfig array to values.yaml: when set, the chart passes the full config inline via --redis.ss.queues-config (sorted-set) or --redis.queues-config (pubsub), bypassing the single-queue CLI flags
- remove usages of ConfigMap or volume mount, the JSON is passed directly as a CLI argument

## Why is this change needed?

The Helm chart was limited to a single queue and single dispatch gate. Users who needed multiple queues with per-queue gates had to create ConfigMaps and manage volume mounts manually outside the chart. This change exposes the full configuration surface that the Go code  already supports.

## How was this tested?

## Checklist

### DIFF
--- a/charts/async-processor/templates/ap-deployments.yaml
+++ b/charts/async-processor/templates/ap-deployments.yaml
@@ -26,26 +26,34 @@ spec:
           {{- if .Values.ap.redis.enabled }}
           {{- if eq (.Values.ap.messageQueueImpl | default "redis-pubsub") "redis-sortedset" }}
           - --message-queue-impl=redis-sortedset
+          {{- if .Values.ap.redis.queuesConfig }}
+          - --redis.ss.queues-config={{ .Values.ap.redis.queuesConfig | toJson }}
+          {{- else }}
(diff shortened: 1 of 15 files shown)


evacchi:
FYI @vishbhat @yizhaodev

shimib (approved):

PR #222, opened 2026-06-04
Introduce Pool Topology

jtechapps:
## What does this PR do?

Inference pool isolation feature from https://github.com/llm-d-incubation/llm-d-async/pull/186

This PR adds a new pool object for grouping and isolating backends.

If queue config file is specified, pool config must also be specified. 
Each queue in the queue config must now specify pool id that should handle requests from the queue.

If a queue config is not specified and only the command line parameters are provided, a single "default" pool will be created.

Multiple queues can specify the same pool id.
The merge policy has been changed to merge channels per pool.
Round robin merge policy will perform round robin balancing for all the queues sharing the same pool id.

## Why is this change needed?

Each pool configures an inference backend and has its own workers. when one pool is saturated it does not block processing other requests on different pools.

## How was this tested?

I tested by running async-processor locally against llm-d-simulator instances running on the same host.

I ran a slow and fast llm-d-simulator:
1. fast `./bin/llm-d-inference-sim --model fake-model --port 8000 --time-to-first-token=1s`
2. slow `./bin/llm-d-inference-sim --model fake-model --port 8001 --time-to-first-token=120s`

I configured two pools with one worker each and two pub/sub queues.

Pool config:
```
[
  {
    "id": "fast-pool",
    "igw_base_url": "http://localhost:8000/",
    "request_path_url": "/v1/completions",
    "workers": 1
  },
  {
    "id": "slow-pool",
    "igw_base_url": "http://localhost:8000/",
    "request_path_url": "/v1/completions"
    "workers": 1
  }
]

```

Pub/sub queue config:
```
[
  {
    "pool_id": "fast-pool",
    "subscriber_id": "projects/${PROJECT}/subscriptions/inference-requests1",
    "gate_type": "constant",
    "gate_params": {}
  },
  {
    "pool_id": "slow-pool",
    "subscriber_id": "projects/${PROJECT}/subscriptions/inference-requests2",
    "gate_type": "constant",
    "gate_params": {}
  }
]
```

Run async-processor with minimal configuration:
```
go run cmd/main.go \
  --message-queue-impl=gcp-pubsub \
  --pubsub.project-id=${PROJECT} \
  --pubsub.result-topic-id=llm-d-async-1-results-1 \
  --pubsub.topics-config-file=pubsub1-config.json \
  --pool-config-file=pool-config1.json \
  --metrics-endpoint-auth=false \
  --metrics-port=19090
```

I added requests to each queue and verified that fast pool was not blocked by stuck workers of the slow pool.

Add requests to queue:
```
# fast queue
gcloud pubsub topics publish llm-d-async-1 \
 --project=${PROJECT} \
 --message='{"id":"q1-req1","created":1700000000,"deadline":2000000000,"endpoint":"/v1/completions","payload":{"model":"fake-model","prompt":"How are you today?"}}'

# slow queue
gcloud pubsub topics publish llm-d-async-2 \
 --project=${PROJECT} \
 --message='{"id":"q2-req1","created":1700000000,"deadline":2000000000,"endpoint":"/v1/completions","payload":{"model":"fake-model","prompt":"What is a dog?"}}'
```

Pull results:
```
gcloud pubsub subscriptions pull projects/${PROJECT}/subscriptions/llm-d-async-1-results-1-sub \
  --project=${PROJECT} \
  --limit=1  --auto-ack
```

## Checklist

## Related Issues

https://github.com/llm-d-incubation/llm-d-async/issues/205


pkg/async/random_robin_policy.go
shimib, line 1: We should add error support in the interface and return an error on unknown pools to avoid typos in tests ending in a random pool selecttion.

jtechapps: Fixed. I added a panic in the MergeRequestChannel setup to crash when it finds a pool id that is not in the config.

cmd/main.go
```diff
@@ -68,6 +69,7 @@ func main() {
 	flag.StringVar(&tlsCert, "tls-cert", "", "Path to client certificate file (PEM) for mTLS")
 	flag.StringVar(&tlsKey, "tls-key", "", "Path to client key file (PEM) for mTLS")
 	flag.BoolVar(&tlsInsecureSkipVerify, "tls-insecure-skip-verify", false, "Skip TLS certificate verification (dev/test only)")
+	flag.StringVar(&poolConfigFile, "pool-config-file", "", "Path to the pools configuration JSON file")
```
evacchi, line 80: I think we'll need to expose this on Helm charts

cmd/main.go
```diff
@@ -102,6 +104,48 @@ func main() {
...
 
 	printAllFlags(setupLog)
+
+	hasQueueConfig := isQueueConfigSet()
+	hasPoolConfig := (poolConfigFile != "")
+
+	if hasQueueConfig != hasPoolConfig {
+		setupLog.Error(fmt.Errorf("both pool-config-file and queues/topics config file must be specified together, or both must be omitted"), "Configuration error")
+		os.Exit(1)
+	}
```
evacchi: this is bit too restrictive, and also I am wondering why we need 2 separate config files instead of one with two sections; for instance (even though it's not necessarily a deal breaker) it would break this PR https://github.com/llm-d-incubation/batch-gateway/pull/458


maybe let's split the config file into 2 sections, and let's make sure that it's clear that "pool" here means **worker** pool (maybe we should use that name in structs/interfaces as well), because otherwise it can be easily mistaken for **inference** pool!

pipeline/pool.go
```diff
@@ -0,0 +1,42 @@
...
+// convention. For destinations that aren't IGW-fronted (external
+// providers, plain HTTP servers, etc.), Pool.ID is a free-form
+// operator label with no IGW meaning.
+type PoolConfig struct {
+	ID             string            `json:"id"`
+	IGWBaseURL     string            `json:"igw_base_url"`
+	RequestPathURL string            `json:"request_path_url"`
+	Workers        int               `json:"workers"`
+	HTTPHeaders    map[string]string `json:"http_headers,omitempty"`
+}
```
evacchi: I am not 100% convinced about bundling routing fields to the worker config. 

If it's a "worker pool", should the routing config stay on the queue side (where it was before) and the worker pool only define concurrency + grouping? 

it would make sense if the worker was tied to a specific endpoint, but routing is handled at the message level (each message carries its own RequestURL and HttpHeaders)

Otherwise you could keep the queue config:
```
  {
    "queue_name": "critical-queue",
    "worker_pool_id": "fast-pool",
    "igw_base_url": "http://fast-model:8000",
    "request_path_url": "/v1/completions",
    "gate_type": "constant"
  }
``` 
and then have an optional worker_pools section:

```
  {
    "id": "fast-pool",
    "workers": 4
  }
```

the default fallback policy could be either 1 single pool for all queues (possibly suboptimal) or one pool per queue (maybe wasteful).

pkg/redis/redisimpl.go
```diff
@@ -150,22 +148,30 @@ type RedisMQFlow struct {
...
 	resultChannel   chan api.ResultMessage
+	pools           []pipeline.PoolConfig
 	drainCancel     context.CancelFunc
 	drainWg         sync.WaitGroup
 	enableTracing   bool
 }
 
-// PubSubOption is a functional option for configuring RedisMQFlow.
-type PubSubOption func(*RedisMQFlow)
+// PubSubRedisOptionOption is a functional option for configuring RedisMQFlow.
```
evacchi:
```suggestion
// RedisOption is a functional option for configuring RedisMQFlow.
```

pkg/async/random_robin_policy.go
```diff
@@ -17,63 +18,85 @@ var _ pipeline.RequestMergePolicy = (*RandomRobinPolicy)(nil)
...
-		return pipeline.EmbelishedRequestChannel{Channel: mergedChannel}
+func (r *RandomRobinPolicy) MergeRequestChannels(channels []pipeline.RequestChannel, pools map[string]pipeline.PoolConfig) pipeline.PoolDispatch {
+	channelsByPool := make(map[string][]pipeline.RequestChannel)
+	for _, ch := range channels {
+		poolID := ch.PoolID
+		if poolID == "" {
+			poolID = "default"
+		}
+		if _, ok := pools[poolID]; !ok {
+			panic(fmt.Sprintf("pool %q not found in pools map", poolID))
```
evacchi: I know this is pre-existing because of the method signature, but this is panicking at runtime, we might want to add some validation of the config if possible and/or an error + log (can be addressed separately)

cmd/main.go
```diff
@@ -122,24 +166,24 @@ func main() {
...
+		flow, err := redis.NewRedisSortedSetFlow(redis.WithGateFactory(gateFactory), redis.WithSortedSetRedisTracing(redisTracing), redis.WithSortedSetPools(pools))
 		if err != nil {
 			setupLog.Error(err, "Failed to create Redis sorted-set flow")
 			os.Exit(1)
 		}
 		impl = flow
 		setupLog.Info("Using Redis sorted-set flow with per-queue gating")
 	case "gcp-pubsub":
-		impl = pubsub.NewGCPPubSubMQFlow()
+		impl = pubsub.NewGCPPubSubMQFlow(pubsub.WithPools(pools))
```
evacchi: this is pre-existing, but it panics instead of returning an error, so we might want to get this aligned to the others. can be addressed separately.

evacchi (commented):
left a few comments, especially around config

cmd/main.go
jtechapps: pool in this case is actually closer to inference pool than worker pool. If IGW is being used then the pool config will be roughly mapped with inference pools.

pipeline/pool.go
jtechapps: The routing config should be in the pool config. This is intentional because when multiple queues are specified and they are using the same endpoint they should specify the same pool because the queues are saturating the same endpoint.

>the default fallback policy could be either 1 single pool for all queues (possibly suboptimal) or one pool per queue (maybe wasteful).

When specifying a single queue via command line then we fall back to creating a single pool. This is backwards compatible as that queue will get all workers. 

If multiple queues are used we should force pools to be used otherwise the fallback logic becomes complicated and we can't make assumptions about what the user wants. The current logic will error if this is the case. 



>it would make sense if the worker was tied to a specific endpoint, but routing is handled at the message level (each message carries its own RequestURL and HttpHeaders)

This makes more sense than putting the endpoint in the queue itself. Because saturation (and per-pool gating in the future) happens independently at the endpoint level. Different endpoints should always be in different pools because saturating one endpoint does not impact a different endpoint.

pkg/async/random_robin_policy.go
jtechapps: It's intended to panic at runtime so that the misconfiguration is immediately surfaced to the user. This happens during when the program first starts and the merge policies are configured in main.go. This is done instead of returning an error because it also forces unit tests use proper configuration. Perhaps os.exit() should be used instead?

pipeline/pool.go
```diff
@@ -0,0 +1,42 @@
...
+func LoadPools(path string) ([]PoolConfig, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	var pools []PoolConfig
+	if err := json.Unmarshal(data, &pools); err != nil {
+		return nil, err
+	}
+	return pools, nil
```
lioraron: `LoadPools` doesn't validate the parsed configs. Duplicate pool IDs would be silently de-duplicated when building `poolsMap` in main.go (last one wins), and empty IDs would produce confusing runtime behavior. Add validation here — at minimum unique non-empty IDs:

```go
seen := make(map[string]bool, len(pools))
for _, p := range pools {
    if p.ID == "" {
        return nil, fmt.Errorf("pool config has empty ID")
    }
    if seen[p.ID] {
        return nil, fmt.Errorf("duplicate pool ID %q", p.ID)
    }
    seen[p.ID] = true
}
```

pkg/redis/redisimpl.go
lioraron: Looks like a find-replace artifact — comment says `PubSubRedisOptionOption` but the type is `RedisOption`.

```suggestion
// RedisOption is a functional option for configuring RedisMQFlow.
```

cmd/main.go
```diff
@@ -181,10 +225,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	totalConcurrency := 0
+	poolsMap := make(map[string]pipeline.PoolConfig)
+	for _, p := range pools {
+		if p.Workers == 0 {
```
lioraron: This checks `== 0` but the worker spawn loop below (line ~265) checks `<= 0`, and the README says "Defaults to the global concurrency setting if omitted or <= 0". A negative value in the config file would slip through here and produce zero workers. Change to `p.Workers <= 0` for consistency, and then the redundant second check can be dropped.

pkg/pubsub/pubsubimpl_test.go
```diff
@@ -119,7 +120,98 @@ func TestProcessMessages_QuotaGating(t *testing.T) {
...
+	}()
+}
+
+func contains(s, substr string) bool {
+	for i := 0; i <= len(s)-len(substr); i++ {
+		if s[i:i+len(substr)] == substr {
+			return true
+		}
+	}
+	return false
```
lioraron: This reimplements `strings.Contains` from the standard library — just use that directly and drop this function.

pipeline/pool.go
lioraron: The pool's purpose is worker isolation and concurrency control (don't let one saturated group block another). I think that routing (where to send) is a separate concern.

The `Worker` function is a generic HTTP dispatcher - it sends to whatever `msg.RequestURL` it receives, regardless of pool. The routing default (base URL, path) logically belongs on the queue, where the request enters the system, with the request's own endpoint always taking precedence.

The argument that - two queues sharing a pool must share the same endpoint - looks to conflate worker grouping with endpoint identity. You could have two queues with different default endpoints sharing the same worker pool - the workers don't care, they just dispatch whatever URL the merge policy resolved.

Keeping the pool config to `{id, workers}` and routing on the queue side also avoids the need for a second config file, which addresses the concern about breaking batch-gateway PR #458.

jtechapps: Workers should be roughly grouped by endpoint identity. If a backend endpoint hangs, it should only block workers in the same pool.

>You could have two queues with different default endpoints sharing the same worker pool - the workers don't care, they just dispatch whatever URL the merge policy resolved.

Is there a use case for this that cannot be achieved via queue gates and (future) pool gates? Why not just create multiple pools and use gates to ensure some workers remain idle? Or implement global concurrency limit to be shared between all workers? IMO The global concurrency limit is a bad idea because it introduces back the problems pool topology is trying to fix.

In the future we could add support for overflow pools could be configured to pull extra capacity from queues if they have extra capacity.

pkg/async/random_robin_policy.go
evacchi: panic is fine, but possibly we might want to have a validation phase where we collect/log errors and then fail at the end. But not relevant to this PR, we can address separately.

pipeline/pool.go
lioraron: @jtechapps Interesting points.

I think that in frequent deployments there would be multiple queues dedicated per model (e.g. each queue is associated with a different workload class), and each request will carry its own router URL and API path. The pool's or queue's base+path url defaults would be typically overridden by the request-level routing.

If we couple routing to the pool, it means either:
- One pool per endpoint combination — which requires arbitrarily many pools, and forces reconfig and redeploy of async-processor whenever a backend adds or changes endpoints. In other words, this adds sensitivity in async-processor - to an aspect that it shouldn't be sensitive to.
- Or the pool's routing fields are dead config that's always overridden.

The pool should be a generic worker isolation primitive — admins decide the granularity of pools (per-model, per-queue, per-workload-class, etc.). Bundling routing into the pool config forces a **specific topology** (one endpoint per isolation group), making a decision that should be the admin's decision.

I think that default routing belongs on the queue, which is typically associated with a specific model and workload class. The pool should be pure worker isolation.
```
// pool config
{"id": "priority-pool", "workers": 8}

// queue config
{"queue_name": "model-a-critical-workload", "pool_id": "priority-pool",
 "igw_base_url": "http://router:8080", "request_path_url": "/v1/chat/completions"}
```

This also avoids needing a second config file (which adds complexity). We can use a single config file — pool definitions (`{id, workers}`) can be an optional `worker_pools` section in the existing queue config. And since the pool is purely about worker isolation, naming it "worker pool" would avoid confusion with the K8s InferencePool CRD.

pkg/redis/redisimpl.go
jtechapps: Done

pkg/pubsub/pubsubimpl_test.go
jtechapps: Done

pkg/async/random_robin_policy.go
jtechapps: This can be addressed in follow up PR because the current pattern that appears in the code base is to panic instead of returning error up the stack and requires refactoring in many places outside of the code changed in this PR.

pipeline/pool.go
jtechapps: Done, added validation and unit test.

cmd/main.go
evacchi: right, I'm just saying that the term "pool" is overloaded here even if there was a 1:1 mapping; so I'd advise being explicit

cmd/main.go
jtechapps: Done, I also added pub sub topics config to helm chart.

jtechapps:
I kept the worker pool config and queue configs as separate files for now to preserve backwards compatibility because topics/queues are stored as top level JSON array. The worker pool config would also have to be loaded in each queue implementation.

jtechapps:
This also fixes https://github.com/llm-d-incubation/llm-d-async/issues/250

Fixes: https://github.com/llm-d-incubation/llm-d-async/issues/250

shimib (approved):

Issue #217, opened 2026-06-02
[Feature]: Metrics


shimib:
### Problem Statement

Currently no comprehensive metrics provided

### Proposed Solution

> **Implementation status legend:** ✅ implemented · 🔄 in review · ⬜ not yet implemented. All implemented metrics live under the `llm_d_async` subsystem and carry `queue_id`, `queue_name`, and `pool_name` labels.

**Traffic Layer - Queue Health & Backlog**

  These metrics tell users how much work is pending and if they are meeting their SLOs (minutes/hours).
- [x] ✅ **Backlog Size (Gauge)**: Total items in each Redis SortedSet. (not easy with PubSub, requires external guide how to integrate) — implemented as `async_broker_backlog` (#244).
Why: Essential for auto-scaling the processor and understanding system pressure.
Note: For GCP PubSub this requires some setup to get this metric from external endpoint as PubSub does not provide this through client API.
- [ ] ⬜ **Deadline Proximity (Histogram)**: The distribution of deadline - now() for items in the queue (requires pre-processing).
Why: Helps identify if a large batch of work is about to "expire" simultaneously.
Note: Will require a pre-processing stage as we don’t look at the messages before consumption currently. Feasible for redis-sortedset only (via the periodic backlog poller + `ZRANGEBYSCORE`); not feasible for PubSub (no peek).
- [x] ✅ **Queue Residence Time (Histogram)**: Time from message ingestion to being pulled by a worker. — implemented as `async_queue_residence_time_millis` (#254).
Why: Measures the actual "async delay" introduced by the system.

**Intelligence Layer - Dispatch Gates & Throttling**

  Since llm-d-async uses gates (Prometheus, Redis, Quota) to protect the inference pool, users need to know why processing is slowing down.
- [x] ✅ **Effective Dispatch Budget (Gauge)**: The value [0.0 - 1.0] currently returned by the gate for each queue. — `async_dispatch_budget` (#290).
- [x] ✅ **Gate Decision Breakdown (Counter)**: Why was a message not pulled?
Labels: reason="gate_closed", reason="quota_exhausted", reason="no_budget". — `async_gate_decisions_total` (#291).
- [x] ✅ **Saturation Metric Value (Gauge)**: If using prometheus-saturation, the raw value being compared against the threshold. — implemented as `async_gate_metric_value` / `async_gate_metric_threshold` (#339).
Why: Helps debug why a gate is closed (e.g., "The pool is at 85% saturation, threshold is 80%").

[...]

PR #254, opened 2026-06-16
feat: add queue residence time metric

shimib:
## What

Adds the **Queue Residence Time** histogram from issue #217, measuring the time a message spends buffered in-process from broker ingestion until a worker pulls it for processing. This isolates the **async delay introduced by the system** (queue time) from model time.

- New metric `llm_d_async_async_queue_residence_time_millis` (Histogram), with the standard `queue_id` / `queue_name` / `pool_name` labels and buckets spanning 1ms–120s.
- Stamped as the message enters the in-process buffer (redis sorted-set and pubsub flows). Retries re-enter via the broker, so each delivery is re-stamped — no stale timestamps across retries.
- Observed at worker pull, guarded on a non-zero ingestion time (drained/untimed messages are skipped).
- Measured in-process, so it is **always registered** (not gated on broker support for publish timestamps).

## Changes

- `api/internal_api.go`: in-process-only `IngestionTime` field on `InternalRequest` (`json:"-"`, never persisted by the custom envelope marshaling).
- `pkg/redis/sortedset_impl.go`, `pkg/pubsub/pubsubimpl.go`: stamp ingestion time as the message enters the in-process channel.
- `pkg/metrics/metrics.go`: define `QueueResidenceTime`, add `RecordQueueResidenceTime`, register the collector.
- `pkg/asyncworker/worker.go`: observe residence time at worker pull.
- Tests: assert the collector is always registered and that a sample is recorded on a successful request.
- `README.md`: document the metric in both metrics tables and add a p95 PromQL example.

## Related

Part of #217.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,7 @@ The processor exports the following Prometheus metrics on the metrics port (defa
 | `llm_d_async_async_shedded_requests_total` | Counter | Rate-limited/shed requests (429) |
 | `llm_d_async_async_message_latency_time_millis` | Histogram | Message latency (Pub/Sub only) |
 | `llm_d_async_async_inference_latency_time_millis` | Histogram | Time spent calling the inference gateway (IGW), isolating model time from queue time |
+| `llm_d_async_async_queue_residence_time_millis` | Histogram | Time a message spent buffered in-process from broker ingestion until a worker pulled it |
 
 **Labels:**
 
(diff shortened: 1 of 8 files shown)


jtechapps:
>1ms–120s

is this too short?

jtechapps (approved):

Issue #256, opened 2026-06-16
[Feature]: gcs_uri multipart/form-data RequestTransform plugin for multi-modal providers


shimib:
## Problem Statement

Multi-modal inference APIs (Whisper-style transcription, OCR, etc.) don't accept the OpenAI-style JSON body the pipeline produces. They expect `multipart/form-data` with a `url` field pointing at a signed object URL. Producers can't put raw media bytes on the broker (messages exceed the size limit), so the queued JSON carries a signed object URL (e.g. a GCS V4 signed URL) in a `gcs_uri` field, and the worker must rewrite the body into multipart right before dispatch — including respecting signed-URL expiry on retries.

This is the first concrete consumer of the request body-transform extension point (#255).

## Proposed Solution

Implement a `RequestTransform` plugin (registered via the framework from #255) that activates on per-message `provider` metadata:

- **Transform:** when the JSON body contains a non-empty `gcs_uri`, emit `multipart/form-data`:
  - write `url=<gcs_uri value>`,
  - pass through the remaining model/request fields as form fields,
  - drop the `gcs_uri` field,
  - reject `file_base64` (not supported on this path) with a fatal invalid-request error,
  - return the multipart `Content-Type` (with boundary) so the worker sets it on the request.
  - if no `gcs_uri` is present, return `handled=false` (default JSON path preserved).
- **Validate (deadline-aware preflight):** parse the signed URL's expiry; if it expires before the message deadline, fail fatally before dispatch so the broker doesn't retry a request that cannot succeed.
- **Config:** registered under its plugin type with a `FactoryFunc`; the matching `provider` value(s) come from plugin parameters (`json.RawMessage`), not hardcoded.

## Alternatives Considered

- **Provider branch in `http_client.go`:** the special-casing this whole effort removes.
- **Hardcoded provider name:** replaced by plugin parameters so multiple providers sharing this shape reuse one plugin.

## Willingness to Contribute

Yes, I can submit a PR.

## Additional Context

- Depends on / blocked by #255.
- Acceptance tests: outgoing request is `multipart/form-data`; includes `url=<signed url>`; carries the other model/request fields; does **not** include a multipart file upload; `file_base64` is rejected; an already-expired (relative to deadline) signed URL fails fatally without a retry.

Issue #255, opened 2026-06-16
[Feature]: Request body-transform extension point (EPP-style plugin framework) for dispatch-time payload rewriting


shimib:
## Problem Statement

The worker assumes one outgoing body shape: the OpenAI-style JSON marshalled from `ReqPayload()`, sent as-is by `HTTPInferenceClient.SendRequest`. Any provider that needs a *different* dispatch-time body (multi-modal APIs that want `multipart/form-data`, providers needing field remapping, signed-URL rewriting, etc.) currently has no choice but to patch a provider-specific branch into `pkg/asyncworker/http_client.go`. That doesn't scale — every such provider becomes a core code change and a permanent special case.

We already have a config-driven extension precedent in the repo (`pipeline.GateFactory` — `CreateGate(gateType, params)`, wired via `WithGateFactory(...)` and constructed in `cmd/main.go`), and EPP has a mature plugin framework we depend on. I'd like to introduce an analogous **request body-transform extension point**: a registry of config-loaded plugins the worker consults at dispatch time, defaulting to a no-op so the existing JSON path is byte-for-byte unchanged when nothing is configured.

This issue tracks the **framework**; concrete plugins (starting with the `gcs_uri` multipart transform) are separate.

## Proposed Solution

Model the extension point on EPP's `pkg/epp/plugins` so it's familiar across the projects:

- **Plugin identity** — embed an EPP-style `Plugin` interface exposing `TypedName() TypedName` (`Type` + `Name`).
- **Typed capability interface** — the transform contract the worker invokes:

```go
// RequestTransform is a dispatch-time body-transform plugin. The worker invokes
// the configured chain after marshalling and before SendRequest. A plugin that
// doesn't recognize a message returns handled=false, preserving the JSON path.
type RequestTransform interface {
    plugins.Plugin // TypedName() plugins.TypedName

    // Validate runs before dispatch. Return a fatal (non-retryable) error to
    // reject the message — e.g. a signed object URL that expires before reqDeadline.
    Validate(payload []byte, metadata map[string]string, reqDeadline int64) error

    // Transform optionally returns a new body and Content-Type for the request.
    Transform(payload []byte, metadata map[string]string) (body []byte, contentType string, handled bool, err error)
}
```

- **Factory + registry** — mirror EPP exactly:

```go
type FactoryFunc func(name string, parameters json.RawMessage, handle Handle) (RequestTransform, error)
func Register(pluginType string, factory FactoryFunc) // into a package-level Registry
[...]

PR #257, opened 2026-06-16
feat: add request body-transform extension point

shimib:
## What

Implements the request body-transform extension point from #255: an EPP-style plugin framework that lets providers rewrite the outgoing HTTP body and `Content-Type` at dispatch time based on per-message metadata, while leaving the default OpenAI-style JSON path **byte-for-byte unchanged** when no plugin is configured.

This is the **framework only**. The first concrete plugin (gcs_uri multipart, #256) builds on it separately.

## Changes

- **`pkg/plugins/`** — k8s-free base framework mirroring EPP's `pkg/epp/plugins`:
  - `Plugin` interface (`TypedName()`), `TypedName{Type,Name}`.
  - `FactoryFunc(name string, parameters json.RawMessage, handle Handle)`, `Registry`, `Register`.
  - Minimal `Handle` (`Context()` + plugin lookup) and generic `PluginByType[P]`.
  - Deliberately omits EPP's `RunnablePlugin` (controller-runtime) and `PodList`/`NamespacedName` (apimachinery) — the only parts that pull in Kubernetes deps and have no analog for in-process request plugins. New packages import stdlib only.
- **`pkg/asyncworker/transform/`** — the extension point:
  - `RequestTransform` typed interface (embeds `plugins.Plugin`) with `Validate(payload, metadata, reqDeadline)` and `Transform(payload, metadata) -> (body, contentType, handled, err)`.
  - `Chain`: `Validate` runs all transforms (first error wins); `Apply` is **first-handled-wins, short-circuit**. A nil `*Chain` is a no-op.
  - `loader.go`: `PluginSpec{Name,Type,Parameters}`, `LoadConfig` (JSON file, mirrors `pipeline.LoadWorkerPools`), `BuildChain` (registry lookup + unique-name/type/interface validation).
- **`pkg/asyncworker/worker.go`** — integration:
  - `Worker(...)` takes a trailing `*transform.Chain` (nil preserves current behavior).
  - `Validate` runs in `validateAndMarshal` before dispatch — a failure is fatal/non-retryable.
  - `Apply` runs immediately before `SendRequest`; when handled, the body is swapped and `Content-Type` set via a **copied** header map (retry-safe). The public `api.InferenceClient` interface is unchanged.
- **`cmd/main.go`** — new `--transform-config-file` flag; empty = nil chain = unchanged behavior.

## Design decisions

- Package layout: `pkg/plugins` (base) + `pkg/asyncworker/transform` (typed interface + loader), mirroring EPP's base-vs-consumer split.
- `Handle` implemented now for FactoryFunc signature parity with EPP.
- Multiple applicable transforms: first-handled-wins, short-circuit (a single final HTTP body can't be produced by two transforms).

## Not touched

No Gate / flowcontrol infrastructure was modified.

## Testing

- `pkg/plugins`: `TypedName`, registry register/lookup, handle add/get, `PluginByType`.
- `pkg/asyncworker/transform`: chain `Validate`/`Apply` semantics (incl. short-circuit), loader + all `BuildChain` error paths.
- `pkg/asyncworker` (worker-level, `-race`): a transform rewrites the real outgoing HTTP body + `Content-Type`; a `Validate` failure yields a fatal error result with **no dispatch and no retry**.
- `go build`, `go vet`, and `golangci-lint` (0 issues) all clean; full non-e2e suite passes.

Part of #255.

### DIFF
--- a/pkg/asyncworker/transform/chain.go
+++ b/pkg/asyncworker/transform/chain.go
@@ -1,0 +1,58 @@
+package transform
+
+// Chain is an ordered set of RequestTransform plugins applied to each outgoing
+// request. A nil *Chain is valid and behaves as a no-op, so callers that have no
+// transforms configured can pass nil and preserve the default JSON path exactly.
+type Chain struct {
+	transforms []RequestTransform
+}
+
+// NewChain returns a Chain over the given transforms, in priority order.
+func NewChain(transforms []RequestTransform) *Chain {
+	return &Chain{transforms: transforms}
+}
+
+// Len reports the number of transforms in the chain.
+func (c *Chain) Len() int {
+	if c == nil {
+		return 0
+	}
+	return len(c.transforms)
+}
+
+// Validate runs every transform's Validate in order and returns the first error.
+// Each transform decides whether it applies to the message; non-applicable
+// transforms return nil. A nil/empty chain validates successfully.
+func (c *Chain) Validate(payload []byte, metadata map[string]string, reqDeadline int64) error {
+	if c == nil {
+		return nil
+	}
+	for _, t := range c.transforms {
+		if err := t.Validate(payload, metadata, reqDeadline); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Apply runs transforms in order and returns the body produced by the first one
+// that handles the message (first-handled-wins, short-circuit): once a transform
+// returns handled=true, its body and Content-Type are returned and no further
+// transforms run. If no transform handles the message, handled is false and the
+// caller should send the original payload unchanged. A nil/empty chain returns
+// handled=false. An error from any transform short-circuits and is returned.
+func (c *Chain) Apply(payload []byte, metadata map[string]string) (body []byte, contentType string, handled bool, err error) {
+	if c == nil {
+		return nil, "", false, nil
+	}
+	for _, t := range c.transforms {
+		body, contentType, handled, err = t.Transform(payload, metadata)
+		if err != nil {
+			return nil, "", false, err
+		}
+		if handled {
+			return body, contentType, true, nil
+		}
+	}
+	return nil, "", false, nil
+}
--- a/pkg/asyncworker/transform/chain.go
+++ b/pkg/asyncworker/transform/chain.go
@@ -1,5 +1,7 @@
 package transform
 
+import "fmt"
+
 // Chain is an ordered set of RequestTransform plugins applied to each outgoing
 // request. A nil *Chain is valid and behaves as a no-op, so callers that have no
 // transforms configured can pass nil and preserve the default JSON path exactly.
@@ -29,7 +31,7 @@ func (c *Chain) Validate(payload []byte, metadata map[string]string, reqDeadline
 	}
 	for _, t := range c.transforms {
 		if err := t.Validate(payload, metadata, reqDeadline); err != nil {
-			return err
+			return fmt.Errorf("transform %s: %w", t.TypedName(), err)
 		}
 	}
 	return nil
@@ -48,7 +50,7 @@ func (c *Chain) Apply(payload []byte, metadata map[string]string) (body []byte, 
 	for _, t := range c.transforms {
 		body, contentType, handled, err = t.Transform(payload, metadata)
 		if err != nil {
-			return nil, "", false, err
+			return nil, "", false, fmt.Errorf("transform %s: %w", t.TypedName(), err)
 		}
 		if handled {
 			return body, contentType, true, nil
--- a/pkg/asyncworker/transform/loader.go
+++ b/pkg/asyncworker/transform/loader.go
@@ -35,8 +35,8 @@ func LoadConfig(path string) ([]PluginSpec, error) {
 
 // BuildChain instantiates the configured transforms in order, registering each
 // instance on handle, and returns the resulting Chain. Plugin names must be
-// non-empty and unique; each type must be registered in plugins.Registry and the
-// instantiated plugin must implement RequestTransform.
+// non-empty and unique; each type must be registered with plugins.Register and
+// the instantiated plugin must implement RequestTransform.
 //
 // An empty spec list returns an empty (no-op) Chain.
 func BuildChain(specs []PluginSpec, handle plugins.Handle) (*Chain, error) {
@@ -55,7 +55,7 @@ func BuildChain(specs []PluginSpec, handle plugins.Handle) (*Chain, error) {
 		}
 		seen[spec.Name] = true
 
-		factory, ok := plugins.Registry[spec.Type]
+		factory, ok := plugins.Lookup(spec.Type)
 		if !ok {
 			return nil, fmt.Errorf("transform plugin %q has unregistered type %q", spec.Name, spec.Type)
 		}
--- a/pkg/asyncworker/transform/loader.go
+++ b/pkg/asyncworker/transform/loader.go
@@ -1,0 +1,77 @@
+package transform
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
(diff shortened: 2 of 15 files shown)


pkg/plugins/registry.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "encoding/json"
+
+// FactoryFunc instantiates a plugin from its instance name, raw JSON parameters
+// (as found in configuration), and a Handle to host-provided context. The
+// signature mirrors EPP so plugin authoring is familiar across the projects.
+type FactoryFunc func(name string, parameters json.RawMessage, handle Handle) (Plugin, error)
+
+// Registry maps a plugin type to its FactoryFunc.
+var Registry = map[string]FactoryFunc{}
```
lioraron: The exported mutable global `Registry` conflicts with the project's "no mutable globals" convention. Any package can read/write/delete entries — tests already do `delete(plugins.Registry, typ)` to clean up.

Consider making the map unexported with accessor functions:
```go
var registry = map[string]FactoryFunc{}

func Register(pluginType string, factory FactoryFunc) { registry[pluginType] = factory }
func Lookup(pluginType string) (FactoryFunc, bool) { f, ok := registry[pluginType]; return f, ok }
```

This also lets you add duplicate-registration detection (`Register` currently silently overwrites) and future-proof for concurrency safety.

pkg/plugins/handle.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// GetAllPluginsWithNames returns all of the known plugins keyed by name.
+func (h *handlePlugins) GetAllPluginsWithNames() map[string]Plugin {
+	return h.plugins
+}
+
+// NewHandle returns a Handle backed by the given context and an empty plugin set.
+func NewHandle(ctx context.Context) Handle {
+	return &handle{
```
lioraron, line 57: `GetAllPluginsWithNames()` returns `h.plugins` by reference — callers can mutate the internal map, bypassing `AddPlugin`.

Moreover, neither `GetAllPlugins()` nor `GetAllPluginsWithNames()` have any callers in this PR. Same for `PluginByType`. Consider removing these three methods until there is an actual consumer — they carry EPP heritage weight that isn't justified yet. If kept, `GetAllPluginsWithNames` must return a copy of the map.

pkg/plugins/handle.go
```diff
@@ -0,0 +1,94 @@
...
+	GetAllPlugins() []Plugin
+
+	// GetAllPluginsWithNames returns all of the known plugins keyed by name.
+	GetAllPluginsWithNames() map[string]Plugin
+}
+
+// handle is the default Handle implementation.
+type handle struct {
+	ctx context.Context
+	HandlePlugins
```
lioraron, line 32: Missing compile-time interface compliance checks. The project convention requires `var _ Interface = (*Impl)(nil)` for all implementations. This PR introduces multiple interface/impl pairs but has none:

```go
var _ Handle = (*handle)(nil)
var _ HandlePlugins = (*handlePlugins)(nil)
```

Same applies to any concrete `RequestTransform` implementations and test fakes.

pkg/asyncworker/transform/chain.go
```diff
@@ -0,0 +1,58 @@
...
+// Each transform decides whether it applies to the message; non-applicable
+// transforms return nil. A nil/empty chain validates successfully.
+func (c *Chain) Validate(payload []byte, metadata map[string]string, reqDeadline int64) error {
+	if c == nil {
+		return nil
+	}
+	for _, t := range c.transforms {
+		if err := t.Validate(payload, metadata, reqDeadline); err != nil {
+			return err
+		}
```
lioraron, line 35: Both `Validate` and `Apply` return errors from individual transforms without wrapping them with the transform's identity. When a chain has multiple plugins, the caller can't tell which one failed.

Consider wrapping:
```go
if err := t.Validate(payload, metadata, reqDeadline); err != nil {
    return fmt.Errorf("transform %s: %w", t.TypedName(), err)
}
```

Same pattern for the error in `Apply` (line 52).

pkg/asyncworker/transform/transform.go
```diff
@@ -0,0 +1,34 @@
...
+type RequestTransform interface {
+	plugins.Plugin
+
+	// Validate runs before dispatch. It returns a non-nil error to reject the
+	// message outright (the worker treats this as fatal/non-retryable) — for
+	// example, a signed object URL that expires before reqDeadline. Plugins that
+	// do not recognize the message must return nil.
+	//
+	// reqDeadline is the request deadline in Unix seconds.
+	Validate(payload []byte, metadata map[string]string, reqDeadline int64) error
```
lioraron, line 27: `reqDeadline int64` (Unix seconds) is error-prone — callers could pass milliseconds or nanoseconds with no compiler help. The worker already has `reqDeadline` as a `time.Time` internally.

Consider `reqDeadline time.Time` instead. The single conversion point (`time.Unix(deadline, 0)`) should happen once in the worker, not be pushed onto every plugin implementor.

shimib: Good point on the type-safety risk, but I'd like to keep int64 here, for two reasons:

  1. It matches the call site. The worker passes r.ReqDeadline() into the chain, which is already int64 Unix seconds — there isn't a time.Time in scope at that
  point to forward. (The time.Time reqDeadline you may be thinking of is a separate local in the dispatch loop, derived from requestTimeout, not the message
  deadline.)
  2. It matches the codebase convention. Deadlines are int64 Unix seconds throughout — Request.ReqDeadline() returns int64, and validateAndMarshal/retryMessage all
  work in those units. Converting to time.Time only at the plugin boundary would make this interface inconsistent with everything around it, and push a
  time.Unix(...)/.Unix() round-trip onto callers either way.

  I've documented the unit explicitly in the doc comment (// reqDeadline is the request deadline in Unix seconds.) to reduce the foot-gun. If you feel strongly
  that the plugin API should standardize on time.Time regardless, I'm happy to do it as a follow-up that converts the deadline representation consistently rather
  than just at this one boundary — but I'd lean against a one-off conversion here.

jtechapps:
While it shouldn't be done in this CL, we should also support response body-transform.

One use case: model overriding will have to do some sort of response transforming.

shimib:
@jtechapps agreed — model overriding (and envelope/field remapping) is a real need on the response side. Keeping it out of this PR as you suggested; tracked as the symmetric counterpart to #255 in #259, to build on this framework once it lands.

pkg/asyncworker/transform/loader.go
```diff
@@ -0,0 +1,77 @@
...
+// LoadConfig reads and parses a transform plugins configuration JSON file (a
+// JSON array of PluginSpec). It does not instantiate plugins; pass the result to
+// BuildChain.
+func LoadConfig(path string) ([]PluginSpec, error) {
+	data, err := os.ReadFile(path) // #nosec G304 -- path from trusted CLI flag
+	if err != nil {
+		return nil, fmt.Errorf("failed to read transform config file: %w", err)
+	}
+
+	var specs []PluginSpec
```
jtechapps: Do we plan on using the same config file for response transformers in the future? The  "transform-config-file" name indicates that response transforms would go in the same file if so there should be a way to distinguish between request and response plugins.

One way is instead of having the request transforms as a top level array [], have them under requestTranforms field.

```
{
  requestTransforms: [],
}
```

That way in the future we can add:

```
{
  requestTransforms: [],
  responseTransforms: [],
}
```

shimib: Good call — done in aa76e78. The config is now a top-level object keyed by direction: {"requestTransforms": [ ... ]}, so we can add "responseTransforms" later without breaking existing files. I also turned on DisallowUnknownFields so a typo (or responseTransforms before it's supported) fails loudly instead of being silently dropped. The response side is tracked in #259.

jtechapps (approved):

PR #260, opened 2026-06-18
feat: add gcs_uri multipart RequestTransform plugin (#256)

### LINKED ISSUE #256: [Feature]: gcs_uri multipart/form-data RequestTransform plugin for multi-modal providers

shimib:
## What

Implements #256: the first concrete request body-transform plugin, `gcs_uri_multipart`, built on the framework from #257.

Multi-modal inference endpoints (Whisper transcription, OCR) expect `multipart/form-data` with a `url` field pointing at a signed object URL, not the OpenAI-style JSON the pipeline produces. Producers can't put raw media bytes on the broker (size limits), so the queued payload carries a signed URL (e.g. a GCS V4 signed URL) in a `gcs_uri` field, and the worker must rewrite the body into multipart right before dispatch — respecting signed-URL expiry on retries.

## Changes

- **`pkg/asyncworker/transform/gcsmultipart/`** — the plugin (registered via `init()`/`MustRegister`):
  - **Activation:** owns a message only when `metadata.provider` matches a configured provider **and** the payload has a non-empty `gcs_uri`; otherwise the default JSON path is preserved byte-for-byte.
  - **Transform:** writes `gcs_uri` as a `url` form field (a plain field, **not** a file upload), passes the remaining payload fields through as form fields, drops `gcs_uri`, and rejects a non-empty `file_base64` as a fatal/non-retryable invalid-request error. Returns the multipart `Content-Type` (with boundary).
  - **Validate (deadline-aware preflight):** parses the signed URL's expiry (V4 `X-Goog-Date` + `X-Goog-Expires`, or V2 `Expires`) and fails fatally before dispatch if it expires at or before the request deadline, so the broker doesn't retry a request that cannot succeed. Best-effort: no parseable expiry → no rejection. **No new dependency** (stdlib `net/url`/`time`).
- **`cmd/main.go`** — blank import to register the built-in plugin.
- **`README.md`** — new "Request Body Transforms" section documenting `--transform-config-file`, the config shape, and the `gcs_uri_multipart` plugin.

## Config

```json
{
  "requestTransforms": [
    { "name": "whisper-multipart", "type": "gcs_uri_multipart", "parameters": { "providers": ["whisper"] } }
  ]
}
```

## Testing

Table-driven unit tests covering the acceptance criteria from #256:

- Outgoing body is `multipart/form-data`; includes `url=<signed url>`; carries the other request fields; does **not** include a multipart file upload; `gcs_uri` is dropped.
- `file_base64` is rejected (fatal); provider mismatch / missing `gcs_uri` → not handled (default JSON path).
- An already-expired (relative to deadline) signed URL fails fatally without a retry; future expiry / no-expiry pass.
- Factory parameter validation; signed-URL expiry parsing (V4/V2/none/malformed); plugin self-registration.

`go build ./...` (incl. `-tags integration`), `go test -count=2 ./pkg/asyncworker/... ./cmd/...`, and `golangci-lint` all pass. (The `test/e2e` suite requires a live cluster and is not run here.)

Closes #256.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,6 +369,18 @@ Transforms are configured with `--transform-config-file`, pointing at a JSON obj
 
 Each entry has a unique `name`, a registered plugin `type`, and opaque `parameters`. Unknown top-level fields are rejected. When the flag is empty, no transforms are loaded and behavior is unchanged.
 
+With the Helm chart, set `ap.transformConfig` to this same object; the chart renders it to a config file and wires `--transform-config-file` automatically:
+
+```yaml
+ap:
+  transformConfig:
+    requestTransforms:
+      - name: "whisper-multipart"
+        type: "gcs_uri_multipart"
+        parameters:
+          providers: ["whisper"]
+```
+
 ### `gcs_uri_multipart` plugin
 
 Rewrites a JSON body into `multipart/form-data` for endpoints that take a signed object URL. Because producers can't put raw media bytes on the broker, the queued `payload` carries a signed URL (e.g. a GCS V4 signed URL) in a `gcs_uri` field.
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ The architecture adheres to the following core principles:
     - [Per-Queue Dispatch Gates](#per-queue-dispatch-gates)
   - [Request Messages and Consumption](#request-messages-and-consumption)
     - [Request Merge Policy](#request-merge-policy)
+  - [Request Body Transforms](#request-body-transforms)
+    - [`gcs_uri_multipart` plugin](#gcs_uri_multipart-plugin)
   - [Retries](#retries)
   - [Observability](#observability)
     - [OpenTelemetry Tracing](#opentelemetry-tracing)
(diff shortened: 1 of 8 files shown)


jtechapps:
LGTM, we should update the helm chart with the transform config file.

shimib:
Thanks! Added Helm chart support in the latest push: a new `ap.transformConfig` value is rendered to `transform-config.json` in the config ConfigMap and passed via `--transform-config-file` (mirroring `worker-pools.json` / `pubsub-topics.json`). Empty by default, so existing installs are unchanged. README documents the Helm path and helm-unittest covers the set/empty cases.

jtechapps (dismissed):

jtechapps (approved):

PR #262, opened 2026-06-18
feat: add Pub/Sub readiness health check

### LINKED ISSUE #246: [Feature]: readiness probe for pub/sub
### Problem Statement

https://github.com/llm-d-incubation/llm-d-async/pull/239#discussion_r3382398505

> Readiness depth is asymmetric across backends. Only `RedisMQFlow` and `RedisSortedSetFlow` implement `pipeline.HealthChecker` (via `rdb.Ping`). For `gcp-pubsub` / `gcp-pubsub-gated`, `PubSubMQFlow` does not implement it, so this type assertion fails and `checker` stays `nil`. As a result `/readyz` for the pub/sub backends only reflects the in-process `ready` flag and never verifies broker/subscription connectivity — a pod with an unreachable Pub/Sub backend will still report `ready`.
> 
> Not blocking — just flagging so the pub/sub readiness limitation is understood and not mistaken for full backend readiness. Pub/Sub `HealthCheck` support will be added in a follow-up.

We should address this

### Proposed Solution

we should make the readiness probe works correctly and add Pub/Sub `HealthCheck` support

### Alternatives Considered

n/a

### Willingness to Contribute

Yes, with guidance

### Additional Context

_No response_

shimib:
## Summary

Fixes #246.

`/readyz` only runs a backend health check when the active `Flow` implements the optional `pipeline.HealthChecker` interface (`cmd/main.go`). Only the Redis flows did (`rdb.Ping`), so for `gcp-pubsub` / `gcp-pubsub-gated` the type assertion failed, `checker` stayed `nil`, and `/readyz` reflected **only** the in-process `ready` flag — a pod with an unreachable Pub/Sub backend still reported `ready`.

This PR makes `PubSubMQFlow` implement `pipeline.HealthChecker` so `/readyz` verifies real broker/subscription connectivity. The `cmd/main.go` wiring picks it up automatically.

## How it works

`HealthCheck` probes each configured request subscription via `SubscriptionAdminClient.GetSubscription`, reusing `Subscriber().String()` for resource-name normalization (handles both short IDs and full resource paths).

### IAM consideration

`GetSubscription` requires `pubsub.subscriptions.get`, which is **not** included in `roles/pubsub.subscriber` (the consume-only role). To avoid falsely marking correctly-configured consumer pods as not-ready, a `PermissionDenied` response is treated as **healthy** — it still proves the broker round-tripped (reachable). Genuine connectivity/config failures (`Unavailable`, `DeadlineExceeded`, `NotFound`, connection errors) mark the pod not-ready, which still fixes the original bug.

## Tests

Introduces the `pstest` in-memory fake (first use in the repo) covering:
- **Healthy** — subscription exists → `nil`
- **Missing subscription** → error (`NotFound`)
- **Broker unreachable** — fake closed before probe → error (the #246 regression guard)

## Notes for reviewers

- Scope is **subscription** connectivity only, not the result *topic* used for publishing. Happy to add a `GetTopic` check on `resultTopicID` if preferred.
- `Start()` still uses the package-global `pubSubClient`; only the new `HealthCheck` uses the struct `client` field (kept minimal). Can migrate `Start` for consistency if desired.
- `go.mod`: `google.golang.org/grpc` promoted to a direct dependency (now imported in tests); pstest transitive deps added via `go mod tidy`.

## Verification

- `go build ./...`
- `make test`
- `golangci-lint run ./pkg/pubsub/...` (0 issues)


pkg/pubsub/pubsubimpl.go
```diff
@@ -223,6 +228,35 @@ func (r *PubSubMQFlow) Characteristics() pipeline.Characteristics {
...
+// subscription is reachable. It backs the /readyz probe: an unreachable Pub/Sub
+// backend (e.g. broker down or network partition) surfaces as a gRPC Unavailable
+// error and marks the pod not-ready.
+func (r *PubSubMQFlow) HealthCheck(ctx context.Context) error {
+	for _, cd := range r.requestChannels {
+		// Subscriber() normalizes both short IDs and full resource paths; reuse
+		// its result as the fully-qualified name for the admin lookup.
+		name := r.client.Subscriber(cd.subscriberID).String()
+		_, err := r.client.SubscriptionAdminClient.GetSubscription(ctx,
+			&pubsubpb.GetSubscriptionRequest{Subscription: name})
```
jtechapps: Why not use [.Exists()](https://pkg.go.dev/cloud.google.com/go/pubsub#Subscription.Exists)?
I don't think Exists() requires additional IAM permissions.

You can use sub.Exists() in the requestWorker() loop to cache the current state of the subscription to a shared data structure and then read that data structure in HealthCheck().

jtechapps (approved):

PR #266, opened 2026-06-23
Unify and refactor gating interface

### LINKED ISSUE #206: [Feature]: Unified Gating Interface
### Problem Statement

Currently there is a split  in Gating interfaces between DispatchGate and AttributeGate. We need a unified strong inteface for gating.

### Proposed Solution

Gating should support: fast fail, fail with retry and continue. 
Gating should support attachment of Release callbacks to suppport stateful gates (e.g., reservation counters).

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

jtechapps:
## What does this PR do?

 - Unifies legacy `DispatchGate` and `AttributeGate` contracts into a single `pipeline.Gate` interface returning `pipeline.Verdict`.
 - Introduces stateful resource release tracking (`AttachRelease`, `Release`, `RollbackReleases`) on `api.InternalRequest` to automate cleanup.
- Integrates `ApplyChain` evaluation and automatic release callbacks in both Redis and Pub/Sub ingestion pipelines.
- Updates all corresponding mock gates, unit tests, and integration test suites.
- Adds a new `local-max-concurrency` gate type with in-process queue concurrency limiting and dynamic budget scaling.

## Why is this change needed?

https://github.com/llm-d-incubation/llm-d-async/issues/206

Gating should support: fast fail, fail with retry and continue.
Gating should support attachment of Release callbacks to suppport stateful gates (e.g., reservation counters).

This change is needed to unblock per-pool gating and support gates label propagation.

## How was this tested?

## Checklist

## Related Issues

Fixes: https://github.com/llm-d-incubation/llm-d-async/issues/206

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -66,8 +66,3 @@ type ResultMessage struct {
 	Routing  InternalRouting   `json:"-"`
 	Metadata map[string]string `json:"-"`
 }
-
-// ResultMessageExpand is an extended ResultMessage that carries additional context.
-type ResultMessageExpand struct {
-	ResultMessage
-}
--- a/api/api.go
+++ b/api/api.go
@@ -66,3 +66,8 @@ type ResultMessage struct {
 	Routing  InternalRouting   `json:"-"`
 	Metadata map[string]string `json:"-"`
 }
+
+// ResultMessageExpand is an extended ResultMessage that carries additional context.
+type ResultMessageExpand struct {
+	ResultMessage
+}
--- a/pipeline/gate.go
+++ b/pipeline/gate.go
@@ -6,26 +6,37 @@ import (
 	"github.com/llm-d-incubation/llm-d-async/api"
 )
 
+// VerdictAction represents the action to take after evaluating admission gates.
+type VerdictAction int
+
+const (
+	// ActionContinue allows the request to be dispatched.
+	ActionContinue VerdictAction = iota
+	// ActionDrop permanently discards the request and optionally returns a result.
+	ActionDrop
+	// ActionRefuse temporarily rejects the request and requests redelivery/re-enqueue.
+	ActionRefuse
+)
+
 // Verdict carries the outcome of a gating decision.
 type Verdict struct {
-	Terminate bool
-	Redeliver bool
-	Result    *api.ResultMessageExpand
+	Action VerdictAction
+	Result *api.ResultMessage
 }
 
 // Continue returns a verdict to proceed.
 func Continue() Verdict {
-	return Verdict{}
+	return Verdict{Action: ActionContinue}
 }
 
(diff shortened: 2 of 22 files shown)


api/api.go
```diff
@@ -66,3 +66,8 @@ type ResultMessage struct {
 	Routing  InternalRouting   `json:"-"`
 	Metadata map[string]string `json:"-"`
 }
+
+// ResultMessageExpand is an extended ResultMessage that carries additional context.
+type ResultMessageExpand struct {
+	ResultMessage
+}
```
evacchi: I think might be a leftover, no new fields are being added here, so you can just use the original ResultMessage

pipeline/gate.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"context"
+
+	"github.com/llm-d-incubation/llm-d-async/api"
+)
+
+// Verdict carries the outcome of a gating decision.
+type Verdict struct {
+	Terminate bool
+	Redeliver bool
```
evacchi: this is essentially encoding 4 states, but `Terminate + Redeliver` does not make sense; it is true that we have constructors for all and only valid states, but I'd suggest to use an enum so you can indicate all the valid states (plus we can further extend in the future if necessary)

pipeline/gate.go
```diff
@@ -0,0 +1,81 @@
...
+	// A value of 1.0 indicates full capacity available (system is idle).
+	// The system always returns a valid value, even in case of internal error.
+	Budget(ctx context.Context) float64
+	// Apply applies the gating logic to a request.
+	Apply(ctx context.Context, msg *api.InternalRequest) (Verdict, error)
+}
+
+// ApplyChain runs a series of gates sequentially. If any gate fails or indicates a non-continue
+// verdict, the chain terminates immediately and rolls back any state acquired by previous gates in the chain.
+func ApplyChain(ctx context.Context, msg *api.InternalRequest, gates []Gate) (Verdict, error) {
```
evacchi, line 56: Since we have this strategy here, it _might_ also make sense to generalize `CascadeMetricSource` as a general "fallback" chain. Not urgent though.

jtechapps: ApplyChain works with short circuit behavior to fail a gate upon the first failure, while CascadeMetricSource skips failures and returns the first success. CascadeMetricSource fallback logic can be generalized in another PR.

api/api.go
jtechapps: Done

pipeline/gate.go
jtechapps: Done, refactored into VerdictAction enum.

shimib (approved):

PR #273, opened 2026-06-23
ci: add missing tools for pre-commit checks

evacchi:
## What does this PR do?

a few checks are not being run on `pre-commit` because some formatting/linting tools are not installed on CI; there are also 2 separate jobs that partially overlap (pre-commit and ci-pr-check): we should consolidate these.

## Why is this change needed?

`pre-commit` run is incomplete and diverges from a local workflow

## How was this tested?

## Checklist


evacchi:
I think now we have consolidated ci-pr-checks into pre-commit so we only have one tool

shimib (approved):

PR #276, opened 2026-06-23
Add pool-level gates

### LINKED ISSUE #208: [Feature]: Pool-Level Admission Control
### Problem Statement

Need Pool-level gating

### Proposed Solution

Build on top of #205 and #206 . Need to add a chain of gates support per Pool.

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

jtechapps:
## What does this PR do?

Introduces a pool-level admission control gating mechanism. This allows worker pools to block/park incoming requests in-memory when capacity is reached, preventing expensive broker nack-and-retry cycles.

Pool Gate VerdictActions
### 1. Continue (`ActionContinue`)
* **Behavior:** The gate permits the request to proceed.
* **Worker Action:** The worker immediately dispatches the request to the target inference gateway.
* **Use Case:** Normal execution when the pool has available capacity.

### 2. Wait / In-Memory Park (`ActionWait`)
* **Behavior:** The gate instructs the worker that the pool is temporarily congested, but the request should be held locally.
* **Worker Action:** The worker parks the request in-memory, sleeps for `50ms`, and then queries the gate again. This loops until the gate returns `ActionContinue` or the request deadline is reached.
* **Use Case:** Local, low-overhead capacity gating. Avoids sending nack/retry signals to the broker (which can cause database/pubsub round-trip overhead), but occupies the worker thread while waiting.

### 3. Refuse / Re-enqueue (`ActionRefuse`)
* **Behavior:** The gate denies admission to the pool and signals that the worker should release the request.
* **Worker Action:** The worker immediately pushes the request back to the broker (via the `retryChannel` with `0` backoff) and returns.
* **Use Case:** High-congestion scenarios or pool saturation where holding the current request in-memory is undesirable (the current request is low priority (and has far deadline) and there may be a higher priority request when the gate opens).

### 4. Drop (`ActionDrop`)
* **Behavior:** The gate permanently rejects the request.
* **Worker Action:** The worker fails the request immediately with a fatal error (`"request dropped by gate"`), writes the failure result to the result queue, and does not retry.
* **Use Case:** Requests where it is better to fast-fail the request than wait, for example, best-effort near-realtime requests.

### 5. Direct Gating Block (`Apply` Blocks Internally)
* **Behavior:** The gate does not return immediately. Instead, the gate's `Apply` method blocks internally (e.g. waiting on a channel semaphore or sync primitive) until capacity becomes available, and only then returns `ActionContinue`.
* **Worker Action:** The worker thread is blocked waiting inside the `gate.Apply(...)` call. Once it returns, the worker immediately dispatches the request.
* **Use Case:** Best suited for simple, single-gate setups (like `LocalConcurrencyGate` in `blocking` mode) where no recalculation of budget is needed during the wait time.
* **Note:** This should **not** be used in composite gates, as blocking inside one inner gate prevents other gates from being evaluated or refreshed.

## Why is this change needed?

Fixes: https://github.com/llm-d-incubation/llm-d-async/issues/208

## How was this tested?

## Checklist

## Related Issues

https://github.com/llm-d-incubation/llm-d-async/issues/208


shimib:
Possible data race / panic on ir.releases (blocking). In worker.go the snapshot + rollback is registered unconditionally, before the pool-gate check:

  snapshot := len(msg.Releases())
  defer msg.RollbackReleases(snapshot)
  if poolGate != nil { ... }

  msg embeds the same *api.InternalRequest the flow owns. Queue-gated flows free it on another goroutine — redis sorted-set via activeReleases→resultWorker calling  ir.Release(), GCP Pub/Sub via defer ir.Release() — and Release() does r.releases = nil. So after the worker sends the result and its deferred RollbackReleases runs, it  races that Release():

  - concurrent read/write of ir.releases (a -race build should flag it), and
  - if Release() (→ releases = nil) wins and snapshot == 1, RollbackReleases runs r.releases[:1] on a nil slice → panic: slice bounds out of range [:1] in an unrecovered  worker goroutine.

  This hits any queue-gate-attaching deployment (redis-quota, queue-level local-max-concurrency) even with no pool gate, since the defer is unconditional. With a pool  gate too, the pool release lives on the same ir and can run twice (worker + flow), double-draining the blocking semaphore.

  The integration tests call WorkerWithGate with a fresh ir and no real flow, so this path isn't exercised — could you add a -race test wiring a worker to a flow with a queue gate?

  Suggested direction: don't share the request's release slice with the flow lifecycle — at minimum move snapshot/defer inside if poolGate != nil, and ideally have the worker hold the pool-gate release locally rather than via AttachRelease/RollbackReleases.

jtechapps:
>msg embeds the same *api.InternalRequest the flow owns. Queue-gated flows free it on another goroutine — redis sorted-set via activeReleases→resultWorker calling ir.Release(), GCP Pub/Sub via defer ir.Release() — and Release() does r.releases = nil. So after the worker sends the result and its deferred RollbackReleases runs, it races that Release():

Looking into how to clean this up but I was able to create a test that can trigger this scenario.

The problem stems from the fact that the processMessage() will send to the resultChannel before the defer statement runs. 

The GoRoutine that was blocked in the queue GoRoutine is then unblocked and able to concurrently access the Releases at the same time.
```
// Wait for the result worker to finish publishing and signal true
result := <-resultsChannel
```

jtechapps:
I refactor the gate releases to be passed and released by the gate caller (pool worker, and queue flow) instead of attaching to the InternalRequest.

shimib (approved):

Issue #284, opened 2026-06-24
[Feature]: unify config across backends


evacchi:
### Problem Statement

currently each backend uses different sets of flags that are partially overlapping. 
 
- Some backends allow config files, some backends have shortcuts to set flags on the CLI (e.g. SortedSet), etc.
- gates are parsed differently across backends (e.g. sorted set uses a custom StringMap; GCP PubSub uses map[string]string; composite gate defines its own local struct)
- config surfaces are fragmented: queue/topic configs, pool configs, runtime params, and gate configs are spread across separate files, CLI flags, and env vars with different loading patterns
- only one transport can be active per `async` instance, but the config flags do not make this explicit

### Proposed Solution

- We introduce a single config flag to set the backend (e.g. `--transport`) and we pass either a `--transport-config [string]` or a `--transport-config-file [path]` + a `--pool-config-file`
- we allow _some_ env variables to override specific keys (e.g. redis url etc) to specify secrets
- we extract shared parameter passing and pass structured data to the backend constructors

### Alternatives Considered

n/a

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

this is a somewhat significant refactor and might be introduce some breakage, so we should be mindful about that and possibly introduce some temporary measure for backwards compatibility if necessary

PR #287, opened 2026-06-25
refactor: cleanup server/runner

evacchi:
## What does this PR do?

Refactor Runner.Run() so that the logic is more straightforward: move some code to sibling funcs, refactor to a unified defer block, share the same context where possible. 

## Why is this change needed?

Runner.Run() is hard to understand and would benefit from a light refactoring, before we move more pieces around.

## How was this tested?

All tests should still pass.

## Checklist

## Related Issues

follow up to #283, related to #284

### DIFF
--- a/pkg/server/runner.go
+++ b/pkg/server/runner.go
@@ -40,7 +40,7 @@ func NewRunner(opts *Options) *Runner {
 	return &Runner{opts: opts}
 }
 
-func (r *Runner) Run(ctx context.Context) error {
+func (r *Runner) Run(ctx context.Context) (err error) {
 	opts := r.opts
 
 	logging.InitLogging(opts.LoggingOptions(), opts.Observability.Verbosity)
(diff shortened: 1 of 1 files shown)


shimib:
The refactor dropped the setupLog.Error(...) logs from loadWorkerPools, loadFlow, initInferenceClient (TLS), and
  loadTransforms — they now just return err, and main.go handles a Run() error with a bare os.Exit(1). So a bad pool/flow/TLS/transform config now exits with no output.
  Since Run has a named return, maybe log once in the existing defer:

  defer func() {
      if err != nil {
          setupLog.Error(err, "runner failed")
      }
      // ...existing shutdown...
  }()

  Nits (non-blocking): loadTransforms should take ctx first; naked return → return nil; the new pure helpers (loadFlow, loadWorkerPools) are worth a quick unit test. The  pool not found removal is fine — the merge policy already panics on an unknown pool.

jtechapps (approved):

PR #300, opened 2026-07-03
Preserve HTTP status and surface non-HTTP failures in async results

### LINKED ISSUE #298: [Feature]: Include HTTP status code and error category in ResultMessage
## Problem

The current `ResultMessage` wire format carries only a raw payload string:

```go
type ResultMessage struct {
    ID      string `json:"id"`
    Payload string `json:"payload"`
}
```

Producers consuming the result queue cannot determine the outcome of a request — success, HTTP error, or non-HTTP failure all look identical.

## Why this matters

In [llm-d-batch-gateway](https://github.com/llm-d/llm-d-batch-gateway) (async dispatch via PR #458), the OpenAI Batch API spec requires:

- **HTTP responses (any status)**: Written to output file with the status code and body preserved
- **Non-HTTP failures** (network, cancel, drop, deadline): Written to error file with an error code

The sync inference path gives full visibility (`StatusCode`, `ResponseBody`, `ErrorCategory`). The async path loses all of this because `ResultMessage` only has `Payload`.

## Proposed change

```go
type ResultMessage struct {
    ID         string `json:"id"`
    StatusCode int    `json:"status_code"`       // actual HTTP status; 0 = no HTTP call made
    Payload    string `json:"payload"`           // response body when StatusCode > 0
    Error      string `json:"error,omitempty"`   // reason when StatusCode == 0
}
```

### Semantics

- `StatusCode > 0` → HTTP call was made. `Payload` contains the response body. Consumer interprets based on status code.
- `StatusCode == 0` → No HTTP call. `Error` contains the reason (e.g. "deadline exceeded", "cancelled", "pool gating dropped request", "network timeout").

### Examples

| Scenario | `StatusCode` | `Payload` | `Error` |
|----------|-------------|-----------|---------|
| Inference success | 200 | `{"choices": [...]}` | "" |
| Rate limited (fatal, retries exhausted) | 429 | `{"error": {"message": "..."}}` | "" |
| Auth failure | 401 | `{"error": {"message": "unauthorized"}}` | "" |
| Deadline exceeded | 0 | "" | "deadline exceeded" |
| Gate drop | 0 | "" | "Pool gating dropped request" |
| Network failure (fatal) | 0 | "" | "Failed to send request to inference: dial tcp: ..." |
| Cancelled (future feature) | 0 | "" | "cancelled" |

## Where to populate

The dispatcher already has this information — it just discards it:

- **Success** (`worker.go` ~line 232): `client.SendRequest()` returns `responseBody` → set `StatusCode: 200`
- **HTTP error (fatal)** (~line 264): `inferenceErr` has status code and response body via `ClientError` → set `StatusCode` and `Payload`
- **Deadline exceeded** (~line 352): already known → `StatusCode: 0, Error: "deadline exceeded"`
- **Gate drop** (~line 136): `ActionDrop` → `StatusCode: 0, Error: "dropped"`

`CreateErrorResultMessage` currently only takes an error string. It would need `statusCode` and `responseBody` parameters (or the original error) to populate the new fields.

## Backward compatibility

New JSON fields with zero-values or `omitempty` are ignored by existing consumers that only unmarshal `ID` and `Payload`. No breaking change.

## Context

- Batch-gateway sync path: `inferClient.Generate()` → returns `(GenerateResponse, ClientError)` with full HTTP context
- Batch-gateway async path: `asyncClient.GetResult()` → returns `GenerateResponse{Response: rawBytes}` only — `ClientError` is always nil
- This makes sync/async behavior parity impossible without structured result metadata

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

j-mok-dev:
## Why is this PR needed?

`ResultMessage` previously carried only `ID` and `Payload`, making it impossible
for consumers (e.g. batch-gateway) to distinguish HTTP successes from HTTP errors
or non-HTTP failures. The OpenAI Batch API contract requires routing HTTP responses
(any status) to the output file and non-HTTP failures to the error file — this
distinction was not expressible in the old wire format.

## What does this PR do?

Adds structured fields to `ResultMessage` so consumers can branch on result type:

- **`StatusCode`** (int): actual HTTP status from upstream; non-zero means "HTTP
  response was received." Zero means non-HTTP failure.
- **`ErrorCode`** (string): machine-readable result-level code for non-HTTP failures
  (`DEADLINE_EXCEEDED`, `GATE_DROPPED`, `GATE_ERROR`, `INFERENCE_ERROR`, `INVALID_REQUEST`).
- **`ErrorMessage`** (string): human-readable description.

Supporting changes:

`InferenceClient.SendRequest` now returns `*InferenceResponse` (struct with
   `StatusCode` and `Body`) so the worker propagates the real HTTP status
   (not hard-coded 200) for success paths.
- On retry exhaustion, the last HTTP status/body is preserved via
  `deadlineExhaustedResult` rather than being collapsed into `DEADLINE_EXCEEDED`.
- Shared constructors (`NewErrorResult`, `NewHTTPResult`, `NewGateDroppedResult`,
  `NewDeadlineExceededResult`) in the `api` package ensure all transports
  (asyncworker, pubsub, redis sorted set) produce identical wire format.

**Breaking changes:**
- `InferenceClient.SendRequest` signature changed from `([]byte, int, error)` to
  `(*InferenceResponse, error)` where `InferenceResponse` is a new struct with
  `StatusCode int` and `Body []byte` fields.
- HTTP error `Payload` now contains the raw upstream response body instead of a
  wrapped error string (e.g. `{"error":"INVALID_REQ: client error: status code 400"}`
  → actual response body from upstream).
- Gate-drop `Payload` changed from `{"status":"dropped"}` to
  `{"error":"Pool gating dropped request"}` (unified via `NewGateDroppedResult`
  across all transports). Consumers should use `ErrorCode == "GATE_DROPPED"` for
  detection rather than parsing the payload body.
- Gate-drop results now include `Metadata` (from `req.ReqMetadata()`), which the
  previous inline construction omitted.

## How was this tested?

- [x] Unit tests added/updated/verified

New test coverage:
- `TestSuccessfulRequest_PreservesActualStatusCode` — verifies non-200 success status propagation
- `TestRetryMessage_deadlineExhaustedPreservesHTTPResponse` — verifies last HTTP response survives retry exhaustion
- `TestSendRequest_bodyReadFailurePreservesStatusCode` — verifies partial HTTP response preserves status
- `TestMarshalResultMessage_StructuredFields` — verifies JSON round-trip of all new fields
- `TestSortedSetFlow_ResultStructuredFields` — end-to-end Redis publish/consume of structured fields

## Checklist

- [x] Pre-commit checks pass (`make pre-commit`)
- [x] Unit tests pass (`make test`)
- [x] Integration tests pass (`make test-e2e`)

## Related Issues

Closes #298

### DIFF
--- a/api/inference_client.go
+++ b/api/inference_client.go
@@ -6,7 +6,12 @@ import "context"
 // This interface allows for pluggable implementations beyond the default HTTP client.
 type InferenceClient interface {
 	// SendRequest sends an inference request to the specified URL with the given headers and payload.
-	// Returns the response body and any error that occurred.
+	// Returns the response body, the HTTP status code, and any error that occurred.
+	//
+	// statusCode is the actual HTTP status code from the upstream response. It is non-zero
+	// whenever an HTTP response was received (including error responses like 4xx/5xx).
+	// A zero statusCode means no HTTP response was obtained (e.g. transport/network failure).
+	//
 	// Errors should implement InferenceError to provide an ErrorCategory via Category().
 	// ErrorCategory determines retry and shedding behavior through its Fatal() and Sheddable() methods:
 	//   - ErrCategoryRateLimit:  retryable, sheddable (e.g. 429)
@@ -16,5 +21,5 @@ type InferenceClient interface {
 	//   - ErrCategoryParse:      not retryable
 	//   - ErrCategoryUnknown:    not retryable
 	// A nil error indicates a successful response.
-	SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) (responseBody []byte, err error)
+	SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) (responseBody []byte, statusCode int, err error)
 }
--- a/api/inference_client.go
+++ b/api/inference_client.go
@@ -2,15 +2,24 @@ package api
 
 import "context"
 
+// InferenceResponse holds the HTTP response from an upstream inference request.
+// StatusCode is non-zero whenever an HTTP response was received (including 4xx/5xx).
+// A zero StatusCode means no HTTP response was obtained (e.g. transport/network failure).
+// Body contains the response body; it may be partial if a read error occurred.
+type InferenceResponse struct {
+	StatusCode int
+	Body       []byte
+}
+
 // InferenceClient defines the interface for sending inference requests.
 // This interface allows for pluggable implementations beyond the default HTTP client.
 type InferenceClient interface {
 	// SendRequest sends an inference request to the specified URL with the given headers and payload.
-	// Returns the response body, the HTTP status code, and any error that occurred.
 	//
-	// statusCode is the actual HTTP status code from the upstream response. It is non-zero
-	// whenever an HTTP response was received (including error responses like 4xx/5xx).
-	// A zero statusCode means no HTTP response was obtained (e.g. transport/network failure).
+	// On success (nil error), the returned *InferenceResponse is always non-nil.
+	// On error, *InferenceResponse may still be non-nil if the upstream sent an HTTP
+	// response (e.g. 4xx/5xx); callers should check resp.StatusCode > 0 to determine
+	// whether an HTTP response was received.
 	//
 	// Errors should implement InferenceError to provide an ErrorCategory via Category().
 	// ErrorCategory determines retry and shedding behavior through its Fatal() and Sheddable() methods:
@@ -20,6 +29,5 @@ type InferenceClient interface {
 	//   - ErrCategoryAuth:       not retryable
 	//   - ErrCategoryParse:      not retryable
 	//   - ErrCategoryUnknown:    not retryable
-	// A nil error indicates a successful response.
-	SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) (responseBody []byte, statusCode int, err error)
+	SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) (*InferenceResponse, error)
 }
--- a/pkg/asyncworker/http_client.go
+++ b/pkg/asyncworker/http_client.go
@@ -25,10 +25,10 @@ func NewHTTPInferenceClient(client *http.Client) *HTTPInferenceClient {
 }
 
 // SendRequest implements InferenceClient for HTTP-based inference requests.
-func (h *HTTPInferenceClient) SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) ([]byte, error) {
+func (h *HTTPInferenceClient) SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) ([]byte, int, error) {
 	request, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewBuffer(payload))
 	if err != nil {
-		return nil, &asyncapi.ClientError{
+		return nil, 0, &asyncapi.ClientError{
 			ErrorCategory: asyncapi.ErrCategoryInvalidReq,
 			Message:       "failed to create request",
 			RawError:      err,
@@ -41,7 +41,7 @@ func (h *HTTPInferenceClient) SendRequest(ctx context.Context, url string, heade
 
 	result, err := h.client.Do(request)
 	if err != nil {
-		return nil, &asyncapi.ClientError{
+		return nil, 0, &asyncapi.ClientError{
 			ErrorCategory: asyncapi.ErrCategoryUnknown,
 			Message:       "failed to send request",
 			RawError:      err,
@@ -51,44 +51,47 @@ func (h *HTTPInferenceClient) SendRequest(ctx context.Context, url string, heade
 
 	body, err := io.ReadAll(result.Body)
 	if err != nil {
-		// Response read errors are retryable as the request may have succeeded
-		return nil, &asyncapi.ClientError{
+		return nil, result.StatusCode, &asyncapi.ClientError{
 			ErrorCategory: asyncapi.ErrCategoryServer,
 			Message:       "failed to read response",
 			RawError:      err,
+			StatusCode:    result.StatusCode,
 		}
 	}
 
 	// Check for rate limiting / load shedding (429)
 	if result.StatusCode == 429 {
 		retryAfter, _ := parseRetryAfter(result.Header.Get("Retry-After"))
-		return body, &asyncapi.ClientError{
+		return body, result.StatusCode, &asyncapi.ClientError{
 			ErrorCategory: asyncapi.ErrCategoryRateLimit,
 			Message:       fmt.Sprintf("rate limited: status code %d", result.StatusCode),
 			RawError:      nil,
 			RetryAfter:    retryAfter,
+			StatusCode:    result.StatusCode,
 		}
 	}
 
 	// Check for client errors (4xx, non-429)
 	if result.StatusCode >= 400 && result.StatusCode < 500 {
-		return body, &asyncapi.ClientError{
+		return body, result.StatusCode, &asyncapi.ClientError{
 			ErrorCategory: asyncapi.ErrCategoryInvalidReq,
 			Message:       fmt.Sprintf("client error: status code %d", result.StatusCode),
 			RawError:      nil,
+			StatusCode:    result.StatusCode,
 		}
 	}
 
 	// Check for server errors (5xx)
 	if result.StatusCode >= 500 && result.StatusCode < 600 {
-		return body, &asyncapi.ClientError{
+		return body, result.StatusCode, &asyncapi.ClientError{
 			ErrorCategory: asyncapi.ErrCategoryServer,
 			Message:       fmt.Sprintf("server error: status code %d", result.StatusCode),
 			RawError:      nil,
+			StatusCode:    result.StatusCode,
 		}
 	}
 
-	return body, nil
+	return body, result.StatusCode, nil
 }
 
 // parseRetryAfter parses a Retry-After header value, which can be either
--- a/pkg/asyncworker/http_client.go
+++ b/pkg/asyncworker/http_client.go
@@ -25,10 +25,10 @@ func NewHTTPInferenceClient(client *http.Client) *HTTPInferenceClient {
 }
 
 // SendRequest implements InferenceClient for HTTP-based inference requests.
-func (h *HTTPInferenceClient) SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) ([]byte, int, error) {
+func (h *HTTPInferenceClient) SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) (*asyncapi.InferenceResponse, error) {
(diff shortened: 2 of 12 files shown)


api/inference_client.go
```diff
@@ -16,5 +21,5 @@ type InferenceClient interface {
 	//   - ErrCategoryParse:      not retryable
 	//   - ErrCategoryUnknown:    not retryable
 	// A nil error indicates a successful response.
-	SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) (responseBody []byte, err error)
+	SendRequest(ctx context.Context, url string, headers map[string]string, payload []byte) (responseBody []byte, statusCode int, err error)
```
evacchi: Notice _this is a breaking change in a public API._

it looks like we are now returning both the StatusCode as a naked `int` and as a field in the error. 

if this breaking change is required, then instead of _adding a new return value_, it's probably worth introducing a struct mimicking Go's HTTP Result API that includes both the body and status code as named fields.

e.g.


```go
type InferenceResponse struct {
    StatusCode int
    Body []byte
}

...

func (h *HTTPInferenceClient) SendRequest(
    ctx context.Context, 
    url string, 
    headers map[string]string, 
    payload []byte) (*InferenceResponse, error) 

```

pkg/redis/sortedset_impl.go
```diff
@@ -483,10 +483,7 @@ func (r *RedisSortedSetFlow) processMessages(ctx context.Context, msgChannel cha
 			if verdict.Result != nil {
 				r.resultChannel <- *verdict.Result
 			} else {
-				r.resultChannel <- api.ResultMessage{
-					ID:      rview.ReqID(),
-					Payload: `{"status": "dropped"}`,
-				}
+				r.resultChannel <- api.NewGateDroppedResult(rview, ir.InternalRouting)
```
evacchi, line 488: wire format change, see below

pkg/pubsub/pubsubimpl.go
```diff
@@ -527,10 +527,7 @@ func (r *PubSubMQFlow) processMessages(ctx context.Context, receive receiveFunc,
 			if verdict.Result != nil {
 				resultMsg = *verdict.Result
 			} else {
-				resultMsg = api.ResultMessage{
-					ID:      body.ID,
-					Payload: `{"status": "dropped"}`,
-				}
+				resultMsg = api.NewGateDroppedResult(&body, ir.InternalRouting)
```
evacchi, line 531: wire format change: gate-dropped payload changed from {"status":"dropped"} to {"error":"..."}, this may break external consumers

pkg/pubsub/pubsubimpl.go
```diff
@@ -527,10 +527,7 @@ func (r *PubSubMQFlow) processMessages(ctx context.Context, receive receiveFunc,
...
 			}
 			resultMsg.Routing = ir.InternalRouting
```
evacchi: this is pre-existing, but it looks like this is overwriting both branches unconditionally, maybe we should drop this line, but let's verify this in a follow-up

evacchi (changes requested):
this PR introduces a public API signature change and a change in the wire format so worth flagging for consumers /cc @shimib 

@j-mok-dev make sure that the e2e tests still run locally before this gets merged

shimib:
Nice change — the `StatusCode > 0` discriminator is a clean way to model the HTTP-vs-non-HTTP split, centralizing construction in the `api` package removes the per-transport divergence, and the test coverage hits the right edge cases (non-200 success, body-read failure, retry-exhaustion preserving the last HTTP body). `make test` passes locally for me.

One thing I'd like documented before merge:

**Undocumented wire-format change for gate-dropped results.** `pkg/pubsub/pubsubimpl.go` and `pkg/redis/sortedset_impl.go` both change the drop payload from `{"status": "dropped"}` to `{"error":"Pool gating dropped request"}` (via `NewGateDroppedResult`, which also sets `ErrorCode: GATE_DROPPED`). The "Breaking changes" section documents the HTTP-error `Payload` change but not this one. No in-repo consumer depends on the old string, but the external batch-gateway might key on `status: dropped` — please add it to the breaking-changes list (and confirm the consumer side is updated).

Two minor, non-blocking notes:

- `pubsubimpl.go`: `NewGateDroppedResult(&body, ir.InternalRouting)` already sets `Routing`, then the following `resultMsg.Routing = ir.InternalRouting` sets it again. Harmless (that line still serves the `verdict.Result != nil` branch), just redundant for the new path.
- The gate-drop result now also carries `Metadata` (from `req.ReqMetadata()`), which the old inline construction did not. Probably desirable, but it is a silent behavior change worth a mention.

pkg/asyncworker/http_client.go
```diff
@@ -51,44 +51,47 @@ func (h *HTTPInferenceClient) SendRequest(ctx context.Context, url string, heade
 
 	body, err := io.ReadAll(result.Body)
 	if err != nil {
-		// Response read errors are retryable as the request may have succeeded
-		return nil, &asyncapi.ClientError{
+		return nil, result.StatusCode, &asyncapi.ClientError{
```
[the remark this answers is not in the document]
j-mok-dev: Fixed — now returns `&InferenceResponse{StatusCode: result.StatusCode, Body: body}` on read error, so any partial bytes from `io.ReadAll` plus the HTTP status code are preserved for downstream consumers.

pkg/asyncworker/worker.go
```diff
@@ -226,17 +226,18 @@ func WorkerWithGate(consumeCtx, requestCtx context.Context, characteristics pipe
...
 						select {
 						case resultChannel <- asyncapi.ResultMessage{
-							ID:       msg.PublicRequest.ReqID(),
-							Payload:  string(responseBody),
-							Routing:  msg.InternalRouting,
-							Metadata: msg.PublicRequest.ReqMetadata(),
+							ID:         msg.PublicRequest.ReqID(),
+							StatusCode: statusCode,
+							Payload:    string(responseBody),
+							Routing:    msg.InternalRouting,
```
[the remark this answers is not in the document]
j-mok-dev: Done — success path now uses NewHTTPResult as well.

pkg/redis/sortedset_impl_test.go
```diff
@@ -453,6 +453,62 @@ func TestSortedSetFlow_ResultFIFO(t *testing.T) {
...
+		{ID: "deadline", Payload: `{"error":"deadline exceeded"}`, ErrorCode: api.ErrCodeDeadlineExceeded, ErrorMessage: "deadline exceeded"},
+		{ID: "gate-drop", Payload: `{"error":"Pool gating dropped request"}`, ErrorCode: api.ErrCodeGateDropped, ErrorMessage: "Pool gating dropped request"},
+	}
+	for _, m := range messages {
+		flow.resultChannel <- m
+	}
+
+	go flow.resultWorker(ctx)
+	time.Sleep(200 * time.Millisecond)
+
```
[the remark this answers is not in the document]
j-mok-dev: Fixed — replaced the fixed sleep with a polling loop that waits until all results are present in Redis (with a 2s timeout).

api/inference_client.go
j-mok-dev: Done — introduced `InferenceResponse` struct with `StatusCode` and `Body` as named fields, exactly as suggested. `SendRequest` now returns `(*InferenceResponse, error)`. The `ClientError.StatusCode` field is retained only for retry logic (`RetryAfter` lookup) internal to the client; the canonical status is carried in the response struct.

pkg/pubsub/pubsubimpl.go
j-mok-dev: Acknowledged — documented this in the PR description under a "Breaking Changes" section:
- Gate-drop payload changed from `{"status":"dropped"}` to `{"error":"Pool gating dropped request"}` (now consistent across all transports via `NewGateDroppedResult`). Consumers should use `error_code: "GATE_DROPPED"` for gate-drop detection rather than parsing the payload body.

j-mok-dev:
> Nice change — the `StatusCode > 0` discriminator is a clean way to model the HTTP-vs-non-HTTP split, centralizing construction in the `api` package removes the per-transport divergence, and the test coverage hits the right edge cases (non-200 success, body-read failure, retry-exhaustion preserving the last HTTP body). `make test` passes locally for me.
> 
> One thing I'd like documented before merge:
> 
> **Undocumented wire-format change for gate-dropped results.** `pkg/pubsub/pubsubimpl.go` and `pkg/redis/sortedset_impl.go` both change the drop payload from `{"status": "dropped"}` to `{"error":"Pool gating dropped request"}` (via `NewGateDroppedResult`, which also sets `ErrorCode: GATE_DROPPED`). The "Breaking changes" section documents the HTTP-error `Payload` change but not this one. No in-repo consumer depends on the old string, but the external batch-gateway might key on `status: dropped` — please add it to the breaking-changes list (and confirm the consumer side is updated).
> 
> Two minor, non-blocking notes:
> 
> * `pubsubimpl.go`: `NewGateDroppedResult(&body, ir.InternalRouting)` already sets `Routing`, then the following `resultMsg.Routing = ir.InternalRouting` sets it again. Harmless (that line still serves the `verdict.Result != nil` branch), just redundant for the new path.
> * The gate-drop result now also carries `Metadata` (from `req.ReqMetadata()`), which the old inline construction did not. Probably desirable, but it is a silent behavior change worth a mention.



Thanks for the review!

1. **Wire-format change**: Will add gate-drop payload change to the breaking-changes list. The batch-gateway consumer side already uses `error_code == "GATE_DROPPED"` for detection (not payload string parsing), so it's compatible.

2. **Redundant Routing overwrite**: Fixed — now only applies to the `verdict.Result != nil` branch.

3. **Metadata inclusion**: Good catch — `NewGateDroppedResult` now passes through `req.ReqMetadata()`, which the old inline literal did not. This is intentional (gate-dropped results should carry the same metadata as other results for consumer routing/correlation), but I'll note it in the breaking-changes section as well.

j-mok-dev:
Verified locally with `make test-e2e` on my Colima setup — 30/30 specs passed.

evacchi (dismissed):

evacchi:
@shimib the change should be fine now, but it's right to note the breaking change, where should @j-mok-dev add the note?

shimib:
We don't have a release-notes home in this repo yet — the router uses release-notes.d/<PR>.md fragments consolidated into  RELEASE-NOTES.md, enforced via its PR template + N+2 deprecation policy. For #300, put it in the PR 'Breaking changes' section now and add a  release-notes.d/unreleased/300.md fragment; I'll open a follow-up to bring the router's release-notes mechanism to llm-d-async so there's a durable place going  forward.
I will create it the fragment

j-mok-dev:
PR description is updated with the breaking-change notes.
It looks like the latest commits require a fresh approval on the current head. I don’t have merge permissions here, so once checks are green and it’s approved again, I’d appreciate a maintainer merge.

shimib (approved):

Issue #261, opened 2026-06-18
Raise the default worker concurrency above 8


shimib:
## Summary

The default worker concurrency is `8`, set in [`cmd/main.go:66`](cmd/main.go#L66):

```go
flag.IntVar(&concurrency, "concurrency", 8, "number of concurrent workers")
```

and mirrored in the Helm chart at `charts/async-processor/values.yaml:21` (`ap.concurrency: 8`).

This default is far too low to saturate any real inference pool and leaves the GPU mostly idle out of the box.

## Why this matters

The processor is I/O-bound: each worker holds one in-flight request and blocks on the Router HTTP call for its full duration. By Little's Law, in-flight concurrency caps throughput:

```
required_workers ≈ target_throughput (req/s) × avg_request_latency (s)
```

Because LLM requests take seconds-to-minutes, `concurrency: 8` caps throughput hard. A measured closed-loop sweep (Llama-3.1-8B on vLLM, single H100-80GB, ~256-token outputs, via the inference gateway + EPP) showed:

| Concurrency | Throughput (req/s) | GPU KV used |
| :--- | :--- | :--- |
| **8 (default)** | **4.2** | <1% (idle) |
| 128 | 35.8 | 12% |
| 192 | 40.4 (knee) | 16% |

The default delivers ~4 req/s and <1% KV utilization — roughly **10× below** what the same hardware sustains at the knee (~40 req/s). Operators who don't tune the flag silently waste ~90% of their GPU.

## Proposal

Raise the default to a value that does meaningful work on a single small GPU while staying within what a single gateway/EPP path handles cleanly (the sweep destabilized a lone replica past ~256). Candidate values for discussion:

- **`128`** — saturates a single small GPU; conservative end of the measured knee.
- **`64`** — safer middle ground if we want headroom for smaller/shared backends.

Whatever value is chosen, the Go flag default (`cmd/main.go`) and the chart default (`values.yaml`) should be updated together so they stay in sync.

## Open questions

- What's the most representative default given the range of backends users run (single small GPU vs. large fleets)? A higher default helps the common "saturate one GPU" case but could surprise users on tiny/shared backends.
- Should we pair the bump with a documentation note pointing operators at the [Async Processor Operations Guide](https://github.com/llm-d/llm-d/blob/main/docs/operations/async-processor.md) for tuning `concurrency` to their own latency/throughput targets?

## References

- `cmd/main.go:66` — flag definition
- `charts/async-processor/values.yaml:21` — chart default
[...]

PR #307, opened 2026-07-13
add generation-aware request cancellation with pre-dispatch drop checks

### LINKED ISSUE #299: [Feature]:  Support per-request cancellation from Producer side
## Summary

Add a `CancelRequests` method to the Producer API that allows a producer (e.g. batch-gateway) to cancel previously submitted requests by their IDs.

## Motivation

In the batch-gateway use case, a user can cancel an entire batch job at any time. When cancellation occurs, the batch-processor needs to cancel individual requests that have already been submitted to the dispatcher. Currently there is no mechanism to:

1. Remove pending requests from the request queue before they are dispatched
2. Signal the dispatcher to drop in-flight requests that are currently being processed
3. Communicate back to the producer that a result was dropped due to cancellation (vs. a normal error)

Without this, the batch-gateway must wait for all submitted requests to complete even after a cancel — wasting inference compute and delaying the cancel acknowledgement to the user.

## Proposed API

```
// Producer interface addition
CancelRequests(ctx context.Context, requestIDs []string) error
```

## Proposed Behavior
Pending requests (still in sorted set): Remove from the request queue so the dispatcher never picks them up. This requires the ability to identify requests by ID in the sorted set (see Design Considerations below).

In-flight requests (already dequeued by dispatcher): The dispatcher should check a cancellation set before forwarding to the backend. If the request ID is in the cancellation set, drop it immediately.

Result signaling: When a request is cancelled (either removed from queue or dropped by dispatcher), the result written to the result queue should indicate cancellation. This depends on structured result support (see #298).

## Design Considerations
Current sorted set structure:
Currently SubmitRequest uses ZADD with the full JSON payload as the member and the deadline as the score:
```
ZADD request-queue <deadline> <full-json-payload>
```
This makes it impossible to ZREM by request ID alone. Possible approaches:

Option A — Change member to request ID, store payload separately:

```
ZADD request-queue <deadline> <request-id>
HSET request-payloads <request-id> <full-json-payload>
```
Dispatcher does ZPOPMIN → gets ID → HGET payload. Cancellation is ZREM + HDEL by ID.

Option B — Cancellation set (no queue structure change):
```
SADD cancel-set <request-id>
```
Dispatcher checks SISMEMBER cancel-set <id> after dequeue. If found, drops the request and writes a cancellation result. Pending requests still get dispatched but are immediately dropped. Simpler to implement but doesn't save queue space.

Option C — Hybrid: Use Option B for immediate implementation, migrate to Option A later for efficiency.

### Atomicity
Whatever approach is chosen, the cancel operation should be atomic or idempotent — cancelling an already-completed or non-existent request should be a no-op, not an error.

### TTL for cancellation set
If using a cancellation set, entries should have a TTL (e.g. aligned with the request deadline) to prevent unbounded growth.

## Expected Integration (batch-gateway side)
When a batch job is cancelled, the batch-processor will:

Stop the submit loop (no more requests sent to dispatcher)
Call CancelRequests(ctx, pendingRequestIDs) for requests already submitted
Continue collecting results from the result queue, marking any returned results as cancelled

## Related Issues
#298 Structured ResultMessage (status code + error fields) — needed to distinguish cancellation results from normal completions

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

CC: @lioraron

j-mok-dev:
## What does this PR do?
This PR adds request cancellation support for the Redis sorted-set async flow and makes cancellation generation-aware so reused request IDs do not inherit stale cancel state.
It:
- adds `CancelRequests()` to the producer and stores an active request token per submission
- writes cancellation markers against the active generation only, then checks them in both dequeue and worker pre-dispatch paths
- emits `CANCELLED` results when a request is dropped before dispatch
- cleans up active/cancel markers safely by token so older results cannot clear newer generations
- adds producer-to-result e2e coverage for the cancel path
## Why is this change needed?
Previously, there was no producer-side API to cancel pending async requests, so requests could still be dispatched after a batch/job had been cancelled.
This change is needed to:
- stop requests before dispatch when cancellation arrives in time
- avoid stale cancellation poisoning when the same request ID is reused
- preserve correctness across races between old results, new submissions, and cancellation markers
Cancellation is best-effort up to dequeue/pre-dispatch. It does not guarantee aborting already in-flight inference requests.
## How was this tested?
Verified with:
- `make test-all`
- `go test ./...` in `api`
- `go test ./...` in `pipeline`
- `go test ./...` in `producer`
- `make test-e2e`
## Checklist
## Related Issues
Fixes #299

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,9 +8,10 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
-# Local replace directives (./api, ./pipeline) must exist before go mod download.
+# Local replace directives (./api, ./pipeline, ./producer) must exist before go mod download.
 COPY api/ api/
 COPY pipeline/ pipeline/
+COPY producer/ producer/
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
--- a/api/api.go
+++ b/api/api.go
@@ -1,6 +1,9 @@
 package api
 
-import "encoding/json"
+import (
+	"context"
+	"encoding/json"
+)
 
 // Request is the public interface for submitting requests to the async queue.
 // It exposes only the caller-visible fields. Concrete types like RequestMessage,
@@ -81,11 +84,27 @@ type ResultMessage struct {
 // These are result-level codes describing why a request could not be completed.
 const (
 	ErrCodeDeadlineExceeded = "DEADLINE_EXCEEDED"
+	ErrCodeCancelled        = "CANCELLED"
 	ErrCodeGateDropped      = "GATE_DROPPED"
 	ErrCodeGateError        = "GATE_ERROR"
 	ErrCodeInferenceError   = "INFERENCE_ERROR"
 	ErrCodeInvalidRequest   = "INVALID_REQUEST"
 )
+
+// CancellationChecker reports whether a specific request generation has been cancelled.
+type CancellationChecker interface {
+	IsCancelled(ctx context.Context, requestID, requestToken string) (bool, error)
+}
+
+// RequestCancellationKey returns the Redis key used for per-request cancellation markers.
+func RequestCancellationKey(requestID string) string {
+	return "request-cancel:" + requestID
+}
+
+// RequestActiveTokenKey returns the Redis key tracking the currently active request generation.
+func RequestActiveTokenKey(requestID string) string {
+	return "request-active:" + requestID
+}
 
 // NewErrorResult builds a non-HTTP error ResultMessage.
 // errorCode must be one of the ErrCode* constants; errMsg is a human-readable description.
@@ -127,4 +146,9 @@ func NewGateDroppedResult(req Request, routing InternalRouting) ResultMessage {
 // NewDeadlineExceededResult builds a ResultMessage for deadline expiry.
 func NewDeadlineExceededResult(req Request, routing InternalRouting) ResultMessage {
 	return NewErrorResult(req, routing, ErrCodeDeadlineExceeded, "deadline exceeded")
+}
+
+// NewCancelledResult builds a ResultMessage for a cancelled request.
+func NewCancelledResult(req Request, routing InternalRouting) ResultMessage {
+	return NewErrorResult(req, routing, ErrCodeCancelled, "cancelled")
 }
--- a/api/internal_api.go
+++ b/api/internal_api.go
@@ -34,6 +34,7 @@ const (
 type InternalRouting struct {
 	RetryCount             int    `json:"retry_count,omitempty"`
 	QueueID                string `json:"queue_id,omitempty"`
+	RequestToken           string `json:"request_token,omitempty"`
 	RequestQueueName       string `json:"request_queue_name,omitempty"`
 	ResultQueueName        string `json:"result_queue_name,omitempty"`
 	TransportCorrelationID string `json:"transport_correlation_id,omitempty"`
--- a/api/internal_api_test.go
+++ b/api/internal_api_test.go
@@ -262,6 +262,9 @@ func assertRouting(t *testing.T, got, want InternalRouting) {
 	if got.RequestQueueName != want.RequestQueueName {
 		t.Errorf("RequestQueueName = %q, want %q", got.RequestQueueName, want.RequestQueueName)
 	}
+	if got.RequestToken != want.RequestToken {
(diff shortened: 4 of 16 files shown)


shimib (commented):
**Hot path: collapse the redundant per-request cancellation checks**

Nice work on the generation/token design — the ID-reuse and cleanup-race handling is solid. One thing I'd like to see before merge is the per-request Redis overhead on the no-cancellation path.

According to this PR each request does ~4 cancellation `GET`s: one at dequeue, then three in the worker — `emitCancelledResultIfNeeded` fires after `RecordAsyncReq` (`worker.go:~90`), again after the pool-gate block (`~194`), and again right before send (`~236`). In the common case (no blocking transform, gate returns `Continue` immediately) there's no work or I/O between those three, so they run microseconds apart and return the same answer — 3 round-trips where 1 would do. Every request pays this even though almost none are ever cancelled, and it's on the hot path of a throughput-oriented dispatcher (more so now that the default concurrency is 64, #261).

The extra checks only earn their keep when a *slow* phase sits between them (a blocking transform, or a gate that parks for seconds). So I'd suggest **time-gating** them — which is exactly the pattern this PR already uses for the gate-wait loop (`nextCancellationCheck := now + cancellationCheckPollInterval`, check only once that passes).

Concretely: hoist a single `nextCancellationCheck` for the whole request and have `emitCancelledResultIfNeeded` no-op unless the interval has elapsed since the last check (share the same stamp the gate loop already advances). Then:
- **Fast path:** the first check fires; the after-gate and before-send checks fall inside the interval and are skipped → **1 worker GET** instead of 3.
- **Slow path:** if a transform or gate actually burned more than the interval, the next check fires — exactly when a cancel could have arrived and it's worth noticing.

The trade-off is negligible and not a real regression in guarantees: in the fast path you might dispatch a request cancelled in the few microseconds between the check and the send — but that race already exists (a cancel can always land right after any check, or mid-flight). Cancellation is documented as best-effort up to dispatch, so this stays within that contract while removing 2–3 Redis reads per request.

j-mok-dev:
> **Hot path: collapse the redundant per-request cancellation checks**
> 
> Nice work on the generation/token design — the ID-reuse and cleanup-race handling is solid. One thing I'd like to see before merge is the per-request Redis overhead on the no-cancellation path.
> 
> According to this PR each request does ~4 cancellation `GET`s: one at dequeue, then three in the worker — `emitCancelledResultIfNeeded` fires after `RecordAsyncReq` (`worker.go:~90`), again after the pool-gate block (`~194`), and again right before send (`~236`). In the common case (no blocking transform, gate returns `Continue` immediately) there's no work or I/O between those three, so they run microseconds apart and return the same answer — 3 round-trips where 1 would do. Every request pays this even though almost none are ever cancelled, and it's on the hot path of a throughput-oriented dispatcher (more so now that the default concurrency is 64, #261).
> 
> The extra checks only earn their keep when a _slow_ phase sits between them (a blocking transform, or a gate that parks for seconds). So I'd suggest **time-gating** them — which is exactly the pattern this PR already uses for the gate-wait loop (`nextCancellationCheck := now + cancellationCheckPollInterval`, check only once that passes).
> 
> Concretely: hoist a single `nextCancellationCheck` for the whole request and have `emitCancelledResultIfNeeded` no-op unless the interval has elapsed since the last check (share the same stamp the gate loop already advances). Then:
> 
> * **Fast path:** the first check fires; the after-gate and before-send checks fall inside the interval and are skipped → **1 worker GET** instead of 3.
> * **Slow path:** if a transform or gate actually burned more than the interval, the next check fires — exactly when a cancel could have arrived and it's worth noticing.
> 
> The trade-off is negligible and not a real regression in guarantees: in the fast path you might dispatch a request cancelled in the few microseconds between the check and the send — but that race already exists (a cancel can always land right after any check, or mid-flight). Cancellation is documented as best-effort up to dispatch, so this stays within that contract while removing 2–3 Redis reads per request.

Good catch. Thanks!
I time-gated the worker cancellation checks with a request-scoped `nextCancellationCheck`, shared across the initial / gate-wait / pre-send call sites. On the fast path this is now a single Redis GET; additional checks only run after the poll interval has elapsed (e.g. long gate wait or slow transform).
Added `TestWorker_FastPathChecksCancellationOnce` to lock that in.

shimib (approved):

j-mok-dev:
@shimib thanks for merging!

Any chance we can tag v0.7.3 (api/producer) soon? I'm blocked on llm-d/llm-d-batch-gateway#543 (CancelRequests consumer). Let me know if there's anything I can help!

shimib:
Sure, will do

PR #321, opened 2026-07-15
Add E2E reddis sorted set queue clearing benchmark

### LINKED ISSUE #309: [Feature]: Add benchmark for queue clearing and saturation
### Problem Statement

There should be E2E tests that function as benchmarks that spin up queue implementation and async-processor instance and generate load and record how long it takes the requests to be serviced and the saturation metrics. This can be used to show and prevent performance regressions and ensure flow control is working.

### Proposed Solution

For the initial benchmark we assume homogenous inference requests:
1. Generate homogenous inference requests and add to queue using producer.
2. Start mock inference backend
3. Start llm-d-async with a given config
4. Record saturation metric and time how long it takes for all the inference requests to be completed.

See the results posted in https://github.com/llm-d-incubation/llm-d-async/issues/304

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

jtechapps:
## What does this PR do?

Adds benchmark that times how long it takes to clear a queue of requests that are identical in size.
Add benchmark for setup using queue gate and setup using pool gate.

## Why is this change needed?

These benchmark tests will fail if it takes longer than 70 seconds to clear the queue of requests. This is to detect regressions in the queue flow handling.

## How was this tested?

## Checklist

## Related Issues

Fixes: #309

## Release note

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_benchmark_test.go
+++ b/test/e2e/e2e_benchmark_test.go
@@ -1,0 +1,257 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"os/exec"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+
+	"github.com/llm-d/llm-d-async/api"
+)
+
+var _ = ginkgo.Describe("Async Processor Performance Benchmark E2E", ginkgo.Ordered, func() {
+	var ctx context.Context
+
+	ginkgo.BeforeAll(func() {
+		redeployEPPWithFlowControl()
+
+		// Patch the simulator to simulate realistic TTFT & ITL
+		// TTFT: 10ms, ITL: 1ms, Max concurrency: 32.
+		// Also increase max waiting queue length so EPP doesn't reject requests.
+		ginkgo.By("Patching llm-sim to simulate realistic latencies")
+		patchSimArgs([]string{
+			"--model", "test-model",
+			"--port", "8000",
+			"--time-to-first-token", "10ms",
+			"--inter-token-latency", "1ms",
+			"--max-num-seqs", "32",
+			"--max-waiting-queue-length", "2000",
+			"--max-model-len", "4096",
+			"--v", "4",
+		})
+	})
+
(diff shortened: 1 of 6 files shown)


shimib:
Thanks for tackling #309! Reviewed the diff; two things before it can go in (validated against current `main`):

## 🔴 Blocking — won't build after merge (module path)
The branch predates #319 (the Go module rename), so `test/e2e/e2e_benchmark_test.go` imports `github.com/llm-d-incubation/llm-d-async/api`, but `main` is now `github.com/llm-d/llm-d-async`. GitHub reports the PR mergeable (it doesn't touch `go.mod`), but I trial-merged it into `main` and the e2e package fails to compile:

```
test/e2e/e2e_benchmark_test.go:16:2: no required module provides package github.com/llm-d-incubation/llm-d-async/api
```

Please rebase onto `main` and update the import to `github.com/llm-d/llm-d-async/api`.

## 🟠 The 70s clearing-time gate looks fragile
`Expect(elapsedTime).To(BeNumerically("<=", 70*time.Second))` is a hard, absolute wall-clock assertion, and by the configured knobs it sits near/below the throughput floor:

- the sim runs `--max-num-seqs 32` (32 concurrent), and each request is ~510ms (`time-to-first-token 10ms` + `max_tokens 500` × `inter-token-latency 1ms`);
- so steady state ≈ 32 / 0.51s ≈ **63 req/s → ~80s for 5000 requests** — before any CI-runner variance (shared runners, arm64 emulation).

Both variants are bounded by the sim's 32, since the async limits are higher (queue `local-max-concurrency: 40`, pool `50`, `concurrency: 64`). Could you share the observed clearing time and margin? If it's this tight I'd suggest making the time **informational** (log, don't fail), using a looser/relative bound (e.g. queue-gate vs pool-gate), or running the benchmark **nightly/opt-in** rather than as a blocking gate.

## 🟠 Runtime cost
Two 5000-request runs, each with a `BeforeEach` sim re-patch + rollout (and `AfterEach` restore + rollout ≈ 4 rollouts total), plus up to 10-min waits each — a sizable addition to the suite. Since both specs use the same realistic-latency sim config, `BeforeAll`/`AfterAll` would halve the rollouts.

shimib (approved):

PR #326, opened 2026-07-16
feat: support Valkey as a drop-in replacement for Redis

### LINKED ISSUE #325: [Feature]: Support Valkey as a drop-in replacement for Redis
## Background

Redis changed its license to RSALv2/SSPL starting with version 7.4 (March 2024), which makes it incompatible with FOSS distribution requirements. This is a blocker for RHOAI adoption, as Red Hat cannot ship or bundle Redis.

Red Hat has already replaced Redis with Valkey across its ecosystem:
- Fedora 41+ obsoletes the Redis package in favor of Valkey ([FESCo change](https://fedoraproject.org/wiki/Changes/Replace_Redis_With_Valkey))
- RHEL 10 does not include Redis; a [migration guide](https://access.redhat.com/solutions/7125256) to Valkey is published

Valkey is a Redis 7.2 fork under the Linux Foundation (BSD 3-Clause). It is wire-protocol compatible with Redis and supported by AWS, Google Cloud, Oracle, Ericsson, and others.

## Current state

llm-d-async already uses `github.com/redis/go-redis/v9` internally, which works with Valkey without code changes. The llm-d ecosystem is already moving toward Valkey support (e.g., [llm-d-kv-cache-manager Valkey example](https://github.com/llm-d/llm-d-kv-cache-manager/blob/main/examples/valkey_example/README.md)). batch-gateway already supports Valkey in configuration, deployment scripts, and CI integration tests.

## Phase 1: Documentation

- Document Valkey as a compatible backend in README
- Add note to CLI flag descriptions that `--redis.url` accepts Valkey URLs
- Add Valkey compatibility note to Helm chart comments in `values.yaml`

## Phase 2: CI validation

- Replace `redis:7-alpine` with `valkey/valkey:8-alpine` in E2E tests, or add a CI matrix to test both backends
- Verify all gate types (`redis`, `redis-quota`) work against Valkey

## Non-goals

- **Renaming `redis.*` CLI flags, Helm values, or package names** — these refer to the wire protocol, not the Redis product. There are 20+ flags/config keys prefixed with `redis.*`; renaming would be a breaking change for all existing deployments.
- **Migrating to `valkey-io/valkey-go` client library** — `go-redis` works for both; this can be revisited later if Valkey-specific features (e.g., RDMA) are needed
- **Removing Redis support** — existing Redis deployments should continue to work

## Related

- [Fedora: Replace Redis With Valkey](https://fedoraproject.org/wiki/Changes/Replace_Redis_With_Valkey)
- [RHEL 10: Redis to Valkey migration](https://access.redhat.com/solutions/7125256)
- [llm-d-kv-cache-manager Valkey example](https://github.com/llm-d/llm-d-kv-cache-manager/blob/main/examples/valkey_example/README.md)

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

j-mok-dev:
## Why is this PR needed?

Redis changed its license to RSALv2/SSPL starting with version 7.4, which is incompatible with FOSS distribution requirements and blocks RHOAI adoption. Red Hat has already replaced Redis with Valkey across Fedora 41+ and RHEL 10. Valkey is a BSD-licensed Redis 7.2 fork under the Linux Foundation that is wire-protocol compatible.

## What does this PR do?

**Documentation:**
- Add a "Backend Compatibility" section to README documenting Valkey as a supported backend
- Add Valkey compatibility notes to Helm chart `values.yaml` comments
- Add Valkey install option to the e2e deployment guide (`docs/guides/e2e-deploy.md`)
- Update `deploy/install.sh` to support `REDIS_CHART` env var (defaults to `bitnami/valkey`, overridable to `bitnami/redis`)

**E2E validation:**
- Switch default E2E image from `redis:7-alpine` to `valkey/valkey:8-alpine`, configurable via `REDIS_IMAGE` env var
- Make `redis.yaml` manifest use `${REDIS_IMAGE}` substitution instead of a hardcoded image
- Fix `kindLoadImage` to use `ctr import` without `--all-platforms`, which fails for multi-arch images when only one platform's layers are present locally (pre-existing issue exposed by the image switch)

**Non-goals (intentional):**
- `redis.*` CLI flags, Helm values, and package names are **not renamed** — they refer to the wire protocol, not a specific product. Renaming would break all existing deployments.
- No migration to `valkey-io/valkey-go` client library — `go-redis` works for both.
- No removal of Redis support — existing Redis deployments continue to work.

## How was this tested?

- [x] `pre-commit run -a` passes (golangci-lint failure is pre-existing on main)
- [x] `go vet ./test/e2e/` passes
- [x] Full E2E suite (36/36 specs) passes with `valkey/valkey:8-alpine` as the message queue backend
- [x] All gate types verified against Valkey: `redis`, `redis-quota`, `prometheus-saturation`, `prometheus-budget`, `composite`

## Checklist

- [x] Pre-commit checks pass (`make pre-commit`)
- [x] Unit tests pass (`make test`)
- [x] E2E tests pass (`make test-e2e`)

## Related Issues

Closes #325

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,7 @@ undeploy-ap-on-k8s:
 #   AP_IMAGE         — async-processor image tag        (default: $(IMAGE_TAG_BASE)/async-processor:e2e-test)
 #   EPP_IMAGE        — EPP image tag                    (default: registry.k8s.io/.../epp:v1.5.0)
 #   SIM_IMAGE        — inference-sim image tag          (default: ghcr.io/llm-d/llm-d-inference-sim:v0.0.0-test)
+#   REDIS_IMAGE      — Redis/Valkey image for E2E MQ    (default: valkey/valkey:8-alpine)
 #   CONTAINER_TOOL   — container runtime                (default: docker)
 #   E2E_SKIP_CLEANUP — set "true" to keep the Kind cluster after tests
 #
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ The architecture adheres to the following core principles:
   - [When to Use](#when-to-use)
   - [Design Principles](#design-principles)
   - [Table of Contents](#table-of-contents)
+  - [Backend Compatibility](#backend-compatibility)
   - [Deployment](#deployment)
   - [Command line parameters](#command-line-parameters)
   - [Dispatch Gates](#dispatch-gates)
@@ -56,6 +57,14 @@ The architecture adheres to the following core principles:
       - [GCP PubSub Command line parameters](#gcp-pubsub-command-line-parameters)
       - [Multiple Topics Configuration File Syntax](#multiple-topics-configuration-file-syntax)
   - [Development](#development)
+
+## Backend Compatibility
+
+The Async Processor uses the Redis wire protocol for its message queue implementations (`redis-sortedset`, `redis-pubsub`) and dispatch gates (`redis`, `redis-quota`). Redis-protocol-compatible backends such as [Valkey](https://valkey.io/) can be used with the existing Redis configuration surface.
+
+All `--redis.*` CLI flags, Helm values, and environment variables (e.g. `REDIS_URL`) work unchanged with Valkey — point them at your Valkey endpoint the same way you would with Redis.
+
+> **Note:** CLI flags and Helm values retain the `redis.*` prefix because it refers to the wire protocol, not a specific product.
 
 ## Deployment
 
@@ -601,7 +610,7 @@ A persisted implementation based on Redis SortedSets.
 ![Async Processor - Redis Sorted Set architecture](/docs/images/redis_sortedset_architecture.png "AP - Redis SortedSet")
 
 #### Redis Sorted Set Command line parameters
-- `redis.url`: Redis URL (e.g. `redis://user:pass@host:port/db` or `rediss://...` for TLS). Can also be set via `REDIS_URL` env var.
+- `redis.url`: Redis/Valkey URL (e.g. `redis://user:pass@host:port/db` or `rediss://...` for TLS). Supports Redis-protocol-compatible backends such as Valkey. Can also be set via `REDIS_URL` env var.
 - `redis.ss.igw-base-url`: Base URL of the IGW (e.g. https://localhost:30800).<br> Mutually exclusive with `redis.ss.queues-config-file` flag.
 - `redis.ss.request-path-url`: Request path url (e.g.: "/v1/completions"). <br> Mutually exclusive with `redis.ss.queues-config-file` flag.")
 - `redis.ss.inference-objective`: InferenceObjective to use for requests (set as the HTTP header x-gateway-inference-objective if not empty).  <br> Mutually exclusive with `redis.ss.queues-config-file` flag.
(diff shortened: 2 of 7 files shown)


shimib:
Nice, well-motivated change — the non-goals (keep `redis.*` naming for the wire protocol, keep `go-redis`, keep Redis support) are exactly right, and I confirmed the branch is current with `main` (module path already `github.com/llm-d/llm-d-async`, so no rebase needed). Two things:

## 🔴 `install.sh` default-Valkey path connects to the wrong service
`deploy_redis()` now defaults `REDIS_CHART=bitnami/valkey`, but the URL is still hardcoded to the **redis** chart's service name:
```sh
local redis_url="redis://${REDIS_RELEASE_NAME}-master.${REDIS_NS}.svc.cluster.local:6379"
```
The bitnami **valkey** chart doesn't create a `<release>-master` service — confirmed via `helm template redis bitnami/valkey`:
```
redis-valkey-primary     ← valkey chart (release "redis")
redis-master             ← redis chart (for comparison)
```
So with the new default the async-processor is pointed at `redis-master…`, which doesn't exist → it can't connect. This isn't caught by the e2e, which uses the custom `test/e2e/yaml/redis.yaml` manifest (Service named `redis`), not the bitnami valkey chart — so 36/36 green validates the Valkey *image*/wire-compat but never exercises this path.

Fix: derive the host from the chart (`…-valkey-primary` for valkey vs `…-master` for redis), or `--set fullnameOverride`/`nameOverride` so the valkey service matches, or template the host from the release.

## 🟡 `kindLoadImage` — note the blast radius
The `docker save … | ctr images import --snapshotter=overlayfs` swap is a fair fix for the `--all-platforms` multi-arch failure, but it now governs **every** image load in the e2e (epp/sim/ap/envoy/redis), not just the MQ image, and it hardcodes the `overlayfs` snapshotter + the `<cluster>-control-plane` node. Fine for standard single-node kind (and the suite passed) — worth a comment noting the broader scope.

Everything else LGTM.

j-mok-dev:
@shimib Thanks for the review!

**🔴 `install.sh` service name mismatch — fixed.**

Good catch. Thanks. Added `--set nameOverride=redis` to `helm upgrade -i` so both `bitnami/redis` and `bitnami/valkey` produce the same `<release>-master` service name, keeping the URL stable regardless of chart selection.

**🟡 `kindLoadImage` blast radius — acknowledged, added comment.**

Agreed. Added a comment noting that this applies to all images (not just the MQ image) and documents the assumptions: single-node Kind cluster with the default control-plane node name and overlayfs snapshotter.

Force-pushed with both fixes. :)

shimib:
Thanks for the quick turnaround! The `kindLoadImage` comment looks good. But I checked the `install.sh` fix with `helm template` and I don't think `nameOverride` closes the gap — `nameOverride` changes the name *prefix*, but the `-master` vs `-primary` *suffix* is hardcoded per chart (redis uses master/replica terminology, valkey uses primary/replica):

```
$ helm template redis bitnami/redis  --set auth.enabled=false --set nameOverride=redis | grep -A1 'kind: Service'
  redis-headless
  redis-master        ← redis
  redis-replicas

$ helm template redis bitnami/valkey --set auth.enabled=false --set nameOverride=redis | grep -A1 'kind: Service'
  redis-headless
  redis-primary       ← valkey (NOT -master)
  redis-replicas
```

`install.sh` builds `redis://${REDIS_RELEASE_NAME}-master.${REDIS_NS}.svc.cluster.local:6379`, and the default chart is now `bitnami/valkey` → so the default install points at `redis-master`, which valkey never creates (it's `redis-primary`). The async-processor can't connect on the default path. (And the e2e won't catch it — it uses the custom `test/e2e/yaml/redis.yaml` Service named `redis`, not the bitnami chart.)

Suggested fix — derive the suffix from the selected chart:

```sh
case "$REDIS_CHART" in
  */valkey) svc="${REDIS_RELEASE_NAME}-primary" ;;
  *)        svc="${REDIS_RELEASE_NAME}-master"  ;;
esac
redis_url="redis://${svc}.${REDIS_NS}.svc.cluster.local:6379"
```

(Both charts do expose a shared `redis-headless` service, but that's headless/multi-pod, so the role-suffixed service is the cleaner target.) Everything else still LGTM.

shimib (approved):

j-mok-dev:
oh you're right. Fixed. Thanks so much!

PR #330, opened 2026-07-20
Refactor CI container image and chart to use main and v0 tags

### LINKED ISSUE #315: [Feature]: Produce nightly AP image
### Problem Statement

we want nightly regression test in llm-d well-lit path gudes

### Proposed Solution

nightly image

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

jtechapps:
## What does this PR do?

 Refactor container image + helm chart tagging on main branch
 refactor the ci-release workflow to do the following:
    - use "main" tag for images built off main branch to follow llm-d-router
      precedent.
    - use "v0" for helm charts built off main branch to follow llm-d-router
      precedent.
    - prefix versions with v similar to llm-d-router versions.

## Why is this change needed?

Fixes: #315

## How was this tested?

## Checklist

## Release note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -83,6 +83,7 @@ jobs:
         if: github.event_name == 'release' || startsWith(github.ref, 'refs/tags/v') || github.ref_name == 'main'
         env:
           VERSION: ${{ steps.tag.outputs.helm_version }}
+          IMAGE_TAG: ${{ steps.tag.outputs.tag }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_ACTOR: ${{ github.actor }}
         run: |
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -35,13 +35,19 @@ jobs:
           COMMIT_SHA=$(echo "${GITHUB_SHA}" | cut -c1-7)
           echo "commit_sha=${COMMIT_SHA}" >> "$GITHUB_OUTPUT"
           if [[ "${GITHUB_EVENT_NAME}" == "release" ]]; then
-            echo "tag=${GITHUB_REF##refs/tags/}" >> "$GITHUB_OUTPUT"
+            TAG="${GITHUB_REF##refs/tags/}"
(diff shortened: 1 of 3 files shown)


shimib (commented):
Thanks for this — the nightly trigger and token/permissions plumbing are sound. One blocking issue before merge, though: **the chart published off `main` references an image tag that never gets built.**

On a main/nightly run the image is built as `:main` (+ `:<sha>`), but the Helm publish runs with `VERSION=helm_version=v0`, and `scripts/publish-helm-chart.sh` does:

```sh
CHART_VERSION="${VERSION#v}"                          # "v0" -> "0"
yq -i '.ap.image.tag = strenv(VERSION)' values.yaml  # ap.image.tag = "v0"
yq -i '.version=CHART_VERSION | .appVersion=VERSION'  # version="0", appVersion="v0"
```

So the main chart pins `ap.image.tag: v0` / `appVersion: v0`, but no `:v0` image is ever pushed (only `:main` and `:<sha>`). `helm install` from that chart → **ImagePullBackOff**. Two coupled problems:

1. **Tag mismatch:** the chart should reference `ap.image.tag: main` (the image the same run built), not `v0`.
2. **`v`-stripping:** the script strips the leading `v`, so the artifact actually publishes as `oci://ghcr.io/llm-d/charts/async-processor:0`, not `:v0` as the PR title intends.

Root cause: `publish-helm-chart.sh` derives *both* the chart version and the image tag from a single `VERSION`, so "chart `v0` → image `:main`" isn't expressible without also editing that script (which this PR doesn't touch).

Suggested fix: decouple them in `publish-helm-chart.sh` — pass the image tag (`main`) separately and set `ap.image.tag=main`, while the chart version can be `v0`/`0.0.0-main`. That way the main chart resolves to an image that actually exists.

jtechapps:
I updated to allow specifying an additional `IMAGE_TAG` environment variable which falls back to `VERSION` to specify `v0` helm version but `main` container version. I also changed all helm versions to be prefixed with `v`. This is to maintain consistent pattern with [llm-d-router](https://github.com/llm-d/llm-d-router/pkgs/container/charts%2Fllm-d-router-standalone/versions?filters%5Bversion_type%5D=tagged)

shimib (approved):

Issue #311, opened 2026-07-13
Gate capacity reservation leaks on retry → inFlight ratchets up → queue wedges permanently


shimib:
### Summary

In the Redis sorted-set flow, the per-queue gate's capacity reservation (a `LocalConcurrencyGate` slot or a `redis-quota` `INCR`) is released **only** when a `ResultMessage` for the request flows back through `resultWorker`. Every **retry** path re-enqueues the request and emits a `RetryMessage` with **no** result, so the reservation is never released. Because the release closure is stored with a plain `activeReleases.Store(reqID, …)`, a later re-dispatch of the same id **overwrites** the prior closure, orphaning it permanently.

Under ordinary retry conditions (a gateway 503, an `ActionRefuse`, an `ActionWait` timeout), `inFlight` climbs monotonically until `Budget()` returns 0, at which point `batchSize` floors to 0 and **the queue stops dispatching entirely**. With `LocalConcurrencyGate` (no TTL) this is permanent; with `redis-quota` it is masked only by the key's TTL reset (itself incorrect — it causes over-admission when the counter resets while requests are still in flight).

### Affected code

- `pkg/redis/sortedset_impl.go:527` — `r.activeReleases.Store(rview.ReqID(), releases)` (plain store; overwrites any existing closure for the id).
- `pkg/redis/sortedset_impl.go:648` — `r.activeReleases.LoadAndDelete(m.ID)` in `resultWorker` — the **only** non-shutdown release site; runs solely on a `ResultMessage`.
- `pkg/asyncworker/worker.go:67,129,169,183,282,427,491` — retry/re-enqueue branches, each emitting a `RetryMessage` and no `ResultMessage` (so no release).
- `pkg/async/inference/flowcontrol/local_concurrency_gate.go:59-61` — `Budget()` returns 0 when `inFlight >= limit`; no TTL on the counter.

### Steps to reproduce (conceptual)

1. Configure a queue with a `local-concurrency` (or `redis-quota` concurrency) gate, `limit = N`.
2. Submit a request whose dispatch hits a retriable condition (inference `503`, pool-gate `ActionRefuse`, or an `ActionWait` timeout).
3. The worker re-enqueues it via `retryChannel` — no result is produced, so `activeReleases[id]` is never released; `inFlight` stays elevated.
[...]

PR #336, opened 2026-07-21
fix: refresh redis-quota concurrency TTL to prevent over-admission (#335)

### LINKED ISSUE #335: Concurrency-mode redis-quota over-admits when the counter TTL expires mid-flight
### Summary

In `RedisQuotaGate` concurrency mode (`pkg/redis/quota_gate.go`, `acquireConcurrency`), in-flight requests are tracked as an `INCR`/`DECR` counter, but the key's `EXPIRE` is set **only when `new_val == 1`** (the first acquire) and is **never refreshed**. The TTL equals `window` (factory default **1 minute**, `gate_factory.go:194`; the multitenant overlays don't set `window`).

Under sustained per-tenant load the counter stays `> 0`, so ~`window` seconds after it first went positive the key **expires mid-flight** → the counter resets to 0 → the next request `INCR`s from 0 → **over-admission up to `limit` extra concurrent** for that tenant, recurring roughly every `window`. It also triggers whenever a single request is in flight longer than `window`. The original epoch's later `DECR`s then hit the fresh key (guarded at `> 0`), under-counting and compounding the over-admission.

This is the correctness residual after #311 / #334: #334 made the counter *accurate* (retries no longer leak `INCR`s), but an accurate-but-positive counter is still wiped by the TTL.

### Fix (minimal, this milestone)

Refresh `EXPIRE` on every acquire (and on release while reservations remain), so the counter expires only after `window` of total inactivity (crash-orphan cleanup) rather than a fixed `window` after the first acquire.

### Follow-up (post-release)

Robust redesign: per-request member tracking (sorted set scored by deadline; acquire counts non-expired members, release `ZREM`s) so the whole counter can never reset and stale members self-expire individually — removes the single-long-request / quiescence edge the minimal fix still allows. Also set a sensible `window` for concurrency-mode quota in the multitenant overlays/guide.

shimib:
## What / why

Fixes **#335** (sibling of #311/#334). In `RedisQuotaGate` concurrency mode (`pkg/redis/quota_gate.go`), the in-flight counter's `EXPIRE` was set **only on the first acquire** (`new_val == 1`) and never refreshed. The TTL = `window` (factory default **1m**; multitenant overlays don't set it), so under sustained per-tenant load the key **expires mid-flight**, the counter resets to 0, and further requests are **admitted beyond `limit`** — recurring every `window`, and for any request in flight longer than `window`.

This is the correctness residual after #334: that PR made the counter *accurate* (retries no longer leak `INCR`s), but an accurate-but-positive counter was still wiped by the TTL.

## Changes (`pkg/redis/quota_gate.go`, `acquireConcurrency`)

- **Refresh `EXPIRE` on every acquire** (drop the `new_val == 1` guard) — the counter can't expire while there's activity; the key now expires only after `window` of **total inactivity** (crash-orphan cleanup).
- **Refresh `EXPIRE` on release** while reservations remain in flight.
- Regression test `TestRedisQuotaGate_Concurrency_TTLRefresh` (TTL refresh across interleaved acquires via `miniredis` `FastForward`) — **fails on `main`, passes with the fix**.

Rate-limit mode is unchanged (its sorted set expires per-member correctly).

## Testing

- `go test ./pkg/redis/...` green (new + existing quota tests).
- Regression test verified to fail on `main` and pass with the fix.
- (The unrelated `producer.TestMultipleTenantsIsolation` flake is pre-existing and in a different, untouched module.)

Closes #335

### DIFF
--- a/pkg/redis/quota_gate.go
+++ b/pkg/redis/quota_gate.go
@@ -109,10 +109,11 @@ func (g *RedisQuotaGate) acquireConcurrency(ctx context.Context, key string) (ap
 		if current and tonumber(current) >= tonumber(ARGV[1]) then
 			return 0
 		end
-		local new_val = redis.call("INCR", KEYS[1])
-		if tonumber(new_val) == 1 then
-			redis.call("EXPIRE", KEYS[1], ARGV[2])
-		end
+		redis.call("INCR", KEYS[1])
+		-- Refresh the TTL on every acquire so the counter cannot expire while
+		-- requests are still in flight (#311 sibling). The key then expires only
+		-- after ARGV[2] seconds of total inactivity (crash-orphan cleanup).
+		redis.call("EXPIRE", KEYS[1], ARGV[2])
 		return 1
 	`
 	// TTL is window size, or a default 5m if window is 0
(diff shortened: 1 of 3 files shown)


pkg/redis/quota_gate.go
```diff
@@ -109,10 +109,11 @@ func (g *RedisQuotaGate) acquireConcurrency(ctx context.Context, key string) (ap
 		if current and tonumber(current) >= tonumber(ARGV[1]) then
```
jtechapps, line 109: should we refresh ttl if key exists?

if you have concurrency limit of 1 then you over admitting can happen often.

jtechapps: Not refreshing on the overflow path is deliberate: it would weaken the crash-orphan reclaim (a counter left high by crashed workers but still getting refused traffic would never self-heal via TTL — the permanent-wedge the TTL prevents).

jtechapps (approved):

PR #340, opened 2026-07-21
feat!: rename Helm chart async-processor → llm-d-async (#179)

### LINKED ISSUE #179: [Feature]: Change container name and chart name to llm-d-async
### Problem Statement

Change container name and chart name to llm-d-async from current async-processor

### Proposed Solution

Rename everywhere

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

shimib:
Completes **#179** — the container image was already `llm-d-async`; this renames the **Helm chart**.

## Changes
- Chart renamed `charts/async-processor` → `charts/llm-d-async`: `Chart.yaml` name, template helper identifiers (`llm-d-async.*`), rendered resource names/labels, dashboard file, and the helm-unittest suites.
- Consumers updated to the new chart path/name: `scripts/publish-helm-chart.sh`, `pre-commit.yml` + `ci-e2e-tests.yaml`, `deploy/install.sh` (+ kind-emulator), `docs/guides/e2e-deploy.md`, and the e2e suite (chart path, release name, resource-name/label selectors; value files renamed too).
- **Bugfix:** `deploy/install.sh` default image repo pointed at the nonexistent `ghcr.io/llm-d/async-processor` → corrected to `ghcr.io/llm-d/llm-d-async`.

## ⚠️ Breaking
The chart now publishes as `oci://ghcr.io/llm-d/charts/llm-d-async` and renders `llm-d-async-*` resources with `app.kubernetes.io/name=llm-d-async`. A Deployment selector is immutable, so **upgrading from an existing `async-processor` install requires uninstall + reinstall**. The old `charts/async-processor` OCI package isn't renamed (GHCR has no redirect) — its last version stays available but gets no new releases.

## Out of scope (intentionally unchanged)
Go binary name, OTEL service name (`async-processor`), and the `*-system` / `*-monitoring` namespace defaults (kept consistent with the Makefile).

## Testing
- `helm lint` ✓ · `helm unittest` **49/49** ✓ · `go vet ./test/e2e/...` ✓ · `go build ./...` ✓.

## Follow-up (separate, cross-repo)
The `llm-d/llm-d` guides reference `charts/async-processor`; they need a follow-up PR to point at `charts/llm-d-async` **after** a release publishes the new package.

Closes #179


charts/llm-d-async/templates/ap-deployments.yaml
```diff
@@ -20,27 +20,27 @@
...
+        {{- include "llm-d-async.selectorLabels" . | nindent 8 }}
     spec:
       securityContext:
         runAsNonRoot: true
         seccompProfile:
           type: RuntimeDefault
       containers:
       - command:
-        - /async-processor
+        - /llm-d-async
```
jtechapps, line 44: change this back or update the docker binary name first https://github.com/llm-d/llm-d-async/blob/main/Dockerfile#L38

charts/llm-d-async/templates/ap-deployments.yaml
```diff
@@ -161,18 +161,18 @@ spec:
           - name: OTEL_TRACES_SAMPLER_ARG
             value: {{ .Values.ap.otel.samplerArg | quote }}
           - name: OTEL_SERVICE_NAME
-            value: "async-processor"
+            value: "llm-d-async"
```
jtechapps, line 165: charts/llm-d-async/templates/ap-deployments.yaml:259 changes:

            - name: OTEL_SERVICE_NAME
              value: "llm-d-async"   # was: "async-processor"

  and the unittest at tests/deployment_test.yaml:449 was updated to match. But the PR body says OTEL service name is intentionally unchanged (async-processor). So
  either:
  - the value change is an accidental over-replace (the updated test masks it), or
  - the intent changed and the description is stale.

charts/llm-d-async/templates/ap-deployments.yaml
shimib: Good catch — fixed in 25d245a. Reverted just the container `command` back to `/async-processor` to match the image's binary (Dockerfile ENTRYPOINT), which I intentionally left unrenamed. My blanket rename under `charts/` had wrongly changed it, and helm-unittest didn't catch it (no assertion on `command`). Left the binary name as-is rather than renaming it — happy to rename the binary too in a follow-up if we want full consistency.

shimib: Update: went with your other option — **renamed the binary** to `llm-d-async` instead of reverting the command. So the Dockerfile now builds/ENTRYPOINTs `llm-d-async` and the chart command is `/llm-d-async` (matches). While doing it I also made the OTEL service name consistent: this branch had already set the chart's `OTEL_SERVICE_NAME` env to `llm-d-async`, but `otel.go`'s default and the e2e Jaeger query still said `async-processor` (would've failed the nightly OTEL e2e) — both now `llm-d-async`, which also matches the existing `otel.go` doc comment. And fixed the Makefile `IMG`/`E2E_IMG` defaults (pointed at the nonexistent `async-processor` image). Namespaces + buildx builder name left as-is. Added the OTEL change to the breaking release note.

jtechapps (approved):

PR #347, opened 2026-07-23
Throttle workers during worker pool message ingestion

### LINKED ISSUE #141: [Feature]: DispatchGates should be assigned to inference targets
### Problem Statement

currently each DispatchGate is assigned to a specific queue

- each queue might be assigned to a different InferenceTarget; 
- however, nothing potentially prevents to assign multiple queues to the _same_ endpoint/InferenceTarget

however, in the latter case, the DispatchGate might erroneously compute the same "dispatch budget" for each input queue; while instead, the budget should be **shared across** them;

e.g.:

- queue 1 targets EP1
- queue 2 targets EP1
- queue 3 targets EP2

if `budget(EP1) = 0.6` and `budget(EP2) = 0.8`:

- queue 3 should be allowed to dispatch 80% of EP2's max capacity
- however, queue 1 and queue 2 should **share** 60% of EP1's max capacity
  - e.g. it could be 30% each (`budget(EP1)/2)`) or any combination such that `budget(queue_1) + budget(queue_2) = budget(EP1)`; for instance, if queue_1 and queue_2 were assigned priorities, then the slice of budget may be defined in terms of such priority; etc...

### Proposed Solution

DispatchGate should not be (just) assigned to queues and instead it should be assigned to inference targets; or, alternatively, some sort of "hierarchy of gates" should be defined, so that the budget can be "shared" across queues.

### Alternatives Considered

for the time being, it's fine to limit 1 queue : 1 gate, as long as the limitation is clearly documented.

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

### LINKED ISSUE #304: [Feature]: Throttle Worker Pool Workers by Gate Budget
### Problem Statement

Currently gate `Budget()` does not throttle the number of active workers in a worker pool it is only used for a binary decision to admit/block/redeliver requests.

If saturation is hit on the worker group then all messages are parked/redelivered, and then when the gate becomes unsaturated, all workers are resumed at the same time which leads to a burst of requests past the saturation.

## Example Test
I was able to create a benchmark which shows that the saturation can exceed the configured saturation due to traffic bursting when all workers are woken up when the endpoint state goes from saturated to unsaturated.

https://github.com/jtechapps/llm-d-async/tree/samplebenchmark

When the gate is on the worker pool (workerPools), the queue gate defaults to wide-open (ConstOpenGate). The queue worker pops a full batch of 64 requests every 50ms and feeds them directly into memory.

*  No Proportional Throttling: Each of the 64 independent worker goroutines evaluates `poolGate.Apply()` as a binary verdict (ActionContinue vs. ActionWait).
*  Parallel Traffic Pump: Because there is no proportional batch-size shrinking, all `64` parallel workers continue firing requests at EPP at full speed (~10ms per request) as long as the last-scraped Prometheus saturation metric remains `< 0.8`. During the Prometheus scrape delay window, those `64` workers effortlessly pump ~174 requests into EPP before the binary switch flips to ActionWait.

When the gate is placed on the queue (queuesConfig), it is evaluated by a single queue worker goroutine before pulling messages from Redis.

*  Proportional Scaling: Instead of a simple yes/no switch, `gate.Budget(ctx)` returns a continuous fraction representing remaining capacity. As EPP saturation approaches 0.8, budget smoothly shrinks toward 0.0.
*  Dynamic Batch Sizes: The queue worker multiplies its batchSize (64) by this budget. As EPP fills up, the queue worker automatically slows down ingestion (e.g., popping 30 requests, then 10, then 2) before saturation ever reaches 0.8. When saturation hits 0.8, batchSize becomes 0, and the worker stops popping messages from Redis entirely.

| Metric / Attribute | Benchmark 1: Queue-Level Gate | Benchmark 2: Worker-Pool-Level Gate |
| :--- | :--- | :--- |
| **Gate Placement** | Message Queue (`queuesConfig`) | Worker Pool (`workerPools`) |
| **Gate Type** | `prometheus-saturation` | `wait-on-refuse` (wrapping `prometheus-saturation`) |
| **Throttling Mechanism** | Proportional ingestion scaling (`batchSize * budget`) | Binary open/close evaluation across 64 parallel workers |
| **Saturation Verdict Action** | `ActionRefuse` (Re-enqueues back to Redis sorted set) | `ActionWait` (Parks goroutine in memory, polling every 50ms) |
| **Total Requests Processed** | `5,000` | `5,000` |
| **Total Queue Clearing Time** | `54.13s` | `1m05.02s` (`65.02s`) |
| **Average Saturation Level** | `0.8118` | `1.7375` |
| **Behavior Under Load** | Smooth ingestion throttling before saturation hits `0.8`; keeps EPP queue stable at target threshold. | Initial burst overshoot due to Prometheus scrape lag across 64 workers, followed by in-memory parking until EPP drains. |

The desired behavior is that both benchmarks should have average saturation around `0.8`.

### Proposed Solution

Implement a mechanism for sleeping/blocking a fraction of workers within a worker pool based on the gates `Budget()` similar to how `Budget()` will control batch size for queue level gates.

### Alternatives Considered

The user can lower the number of a workers so that saturation cannot be overshot. This will work assuming homogenous requests, and no interference from other systems.

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

jtechapps:
## What does this PR do?

This PR throttles request message ingestion into a worker pool from the merged request channel by sleeping workers based on the admission gate Budget().

- added worker ids for worker pools.
- use pool gate Budget to calculate the number of active workers before pulling a request from the request channel for a worker pool
- throttled high-ID workers (workerID > activeLimit) sleep in the polling loop without popping messages from memory channels.
- set async_dispatch_budget{pool_name=...} metric for worker pool gate saturation.
- at least 1 active worker when budget is 0 so that tier priority gate can drop/redeliver any non-reserved interactive or batch messages.

## Why is this change needed?

Fixes: #304
Fixes: #141

## How was this tested?

## Checklist

## Release note

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ The architecture adheres to the following core principles:
   - [Deployment](#deployment)
   - [Command line parameters](#command-line-parameters)
   - [Dispatch Gates](#dispatch-gates)
+    - [Difference between Queue and Pool Gates](#difference-between-queue-and-pool-gates)
+    - [Worker Pool Gate Budget & Proportional Throttling](#worker-pool-gate-budget--proportional-throttling)
     - [Per-Queue Dispatch Gates](#per-queue-dispatch-gates)
   - [Request Messages and Consumption](#request-messages-and-consumption)
     - [Request Merge Policy](#request-merge-policy)
@@ -139,6 +141,17 @@ The Async Processor supports dispatch gates to control batch processing based on
 
 * **Queue-level gates** run at the admission phase for a specific queue. When a queue-level gate denies admission (returning `ActionRefuse`), the request is immediately returned to the broker to be retried/re-delivered, freeing the worker to process other queues.
 * **Pool-level gates** run directly inside the worker loop to regulate capacity constraints shared by all queues routing to that worker pool. When a pool-level gate returns `ActionWait`, the worker parks in-memory and polls until capacity is available, avoiding broker nack/retry overhead. If the pool-level gate returns `ActionRefuse`, the request is immediately returned to the broker.
+
+### Worker Pool Gate Budget & Proportional Throttling
+
+When a pool-level gate is configured, `poolGate.Budget(ctx)` dynamically calculates the available capacity budget $B \in [0.0, 1.0]$ for the worker pool. Rather than treating gate capacity as a binary on/off switch, the Async Processor uses `Budget()` for **Proportional Worker Pool Throttling**:
+
+1. **Active Worker Calculation**: The pool dynamically scales active worker concurrency $A$ based on total pool size $W$ and current budget $B$:
+   $$A = \max\left(1, \lfloor W \times B \rfloor\right) \quad \text{(when } W \ge 1\text{)}$$
+2. **Pre-Pop Throttling (Starvation Prevention)**: Each worker goroutine $w \in \{1 \dots W\}$ evaluates $w \le A$ **before** popping a request from the pool's dispatch channel:
+   - **Active workers ($w \le A$)**: Pull messages from the channel and evaluate `poolGate.Apply()` to dispatch inference requests.
+   - **Throttled workers ($w > A$)**: Sleep in a polling loop without pulling messages, preserving queue FIFO order and preventing head-of-line blocking / request starvation.
+3. **Prometheus Metrics**: The pool budget is exported to Prometheus via `async_dispatch_budget{pool_name="..."}`.
 
 ### Per-Queue Dispatch Gates
 
--- a/docs/dispatch-budget.md
+++ b/docs/dispatch-budget.md
@@ -68,9 +68,26 @@ $$F = 0.3, \quad B = 0.1$$
 2. `max_concurrency` is expressed in request units here, but the same approach generalizes to other capacity units (e.g., tokens, bytes) given an equivalent per-endpoint capacity constant.
 3. We may also define system capacity in terms of number of bytes; in this case $D$ would be truncated at the boundary of a valid request; for instance, if we expect to be able to forward a maximum of 1024 KiB of data, then we may forward at most $1024 \times (0.7-0.1) = 614.4$ KiB; which means if we have $r_1, r_2, ...$ enqueued; if $r_1$ is 600 KiB and $r_2$ is 100 KiB, then we would only dispatch $r_1$
 
+### Worker Pool Budget Enforcement & Proportional Throttling
+
+When applied as a **Pool-Level Gate**, the dispatch budget $D \in [0, 1]$ directly regulates worker concurrency across the pool of $W$ total worker goroutines:
+
+1. **Active Worker Formula**:
+   The active worker count $A$ is dynamically calculated from total pool size $W$ and current budget $D$:
+   $$A = \max\left(1, \lfloor W \times D \rfloor\right) \quad (\text{when } W \ge 1\text{)}$$
+
+2. **Pre-Pop Capacity Guard (Starvation & Burst Prevention)**:
+   Worker goroutine $w \in \{1 \dots W\}$ checks $w \le A$ **before** popping a request from the pool's dispatch channel (`requestChannel`):
+   - **Active Workers ($w \le A$)**: Pull messages from `requestChannel` and evaluate `poolGate.Apply()` to dispatch inference requests to the gateway.
+   - **Throttled Workers ($w > A$)**: Sleep in a polling loop without pulling messages. This prevents head-of-line blocking and request starvation while maintaining queue backpressure on the broker reader.
+
+3. **Metrics Integration**:
+   The active dispatch budget $D$ is exported to Prometheus via the gauge:
+   `async_dispatch_budget{queue_id="", queue_name="", pool_name="<poolID>"}`
+
 ### Failure Modes
 
 * **Queue Consumer Failures:** If the Async Processor pod crashes, the messages remain in the persistent Message Queue (Redis/PubSub), ensuring no data loss.
-* **Downstream Router/IGW Backpressure:** If `llm-d-router` (or another inference gateway) returns an HTTP error indicating overload (e.g. HTTP 429\) despite the computed budget, then the **saturation is assumed to be close to 1**, and therefore a **dispatch budget close to 0\.** The Async Processor will not retry until a subsequent update from the metrics store will bring it back within the acceptable limits.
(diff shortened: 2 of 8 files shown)


jtechapps:
I will add a release note. I am also testing this now using queue clearing benchmark to ensure throughput is maintained.

jtechapps:
I have been testing this and it doesn't fully fix the problems and introduces a few additional issues.

Here are the **issues, edge cases, and architectural failure modes** with the current implementation, including why a **burst of requests past saturation can still occur** and several subtle bugs introduced by the pull-loop throttling logic.

---

### 1. TOCTOU & Concurrency Bugs (Why Bursts Past Saturation Still Occur)

#### A. Stale Eligibility for Workers Parked on an Empty `requestChannel` (The TOCTOU Problem)
* **Location:** `pkg/asyncworker/worker.go:68-81`
* **Why it happens:**
  1. During an idle period or traffic lull when the budget is high (e.g., $D = 1.0$), all $W=64$ workers pass the `workerID <= activeLimit` check (`64 <= 64`).
  2. All 64 workers exit the budget loop and block on `case msg := <-requestChannel:` inside the `select` block.
  3. While these workers are parked waiting for messages on `requestChannel`, suppose cluster saturation spikes and the budget drops to $D = 0.10$ ($A = 6$).
  4. Because workers **do not re-evaluate `Budget(ctx)` while waiting on `<-requestChannel`**, they retain their stale "eligible" status indefinitely.
* **The Burst Edge Case:** As soon as a batch of requests arrives in `requestChannel`, **all 64 parked workers dequeue messages simultaneously**. When they subsequently call `poolGate.Apply()`, gates like `MetricDispatchGate` check `if g.Budget(ctx) <= 0.0`. Because $D = 0.10 > 0.0$, `Apply()` returns `VerdictContinue` for all 64 messages, **causing an immediate burst of 64 requests past saturation**—the exact burst issue this change was intended to prevent.

#### B. 100% CPU Busy-Loop Redelivery Storm when `budget == 0.0` with `Refuse()` Gates
* **Location:** `pkg/asyncworker/worker.go:40-42` (`if budget <= 0.0 { return 1 }`) combined with non-`WaitOnRefuse` gates.
* **Why it happens:**
  1. Notice in `test/e2e/helm/benchmark-pool-gate.yaml`, `gate_type` was changed from `wait-on-refuse` to `prometheus-saturation`.
  2. When saturation reaches 100% ($D = 0.0$), `computeActiveWorkers` returns `1`, allowing Worker 1 (`workerID = 1`) to bypass the sleep loop and dequeue `msg := <-requestChannel`.
  3. Worker 1 calls `poolGate.Apply()`. For `MetricDispatchGate` (and any `composite` gate containing it), `Apply()` returns **`pipeline.Refuse()`** when `Budget(ctx) <= 0.0`.
  4. Worker 1 handles `ActionRefuse` (`lines 202-211`) by sending `msg` to `retryChannel` with `BackoffDurationSeconds: 0`, returning from `processMessage`, and immediately looping back to the top of `WorkerWithGate`.
* **The Infinite Loop Edge Case:** Because `workerID = 1 <= 1` remains true, Worker 1 immediately dequeues the next message from `requestChannel`, refuses it, and dumps it into `retryChannel` with 0 delay. Worker 1 enters a **100% CPU infinite spin loop**, continuously cycling all queued messages through `retryChannel` $\rightarrow$ `requestChannel` without sleeping.
  * *Note:* The integration test (`TestPoolGating_ProportionalThrottling`) did not catch this because its mock `dynamicBudgetGate` returns `Wait()` when `budget <= 0`, whereas real `prometheus-saturation` gates return `Refuse()`.

#### C. Head-of-Line (HOL) Blocking and Queue Paralysis when `budget == 0.0` with `Wait()` Gates
* **Location:** `worker.go:40-42` combined with `TierPriorityAdmissionGate` or `WaitOnRefuseGate`.
* **Why it happens:** When $D = 0.0$, `activeLimit` is `1`. Only Worker 1 is permitted to pull messages from `requestChannel`; Workers $2 \dots W$ are sleeping.
* **The HOL Edge Case:**
  * In a `TierPriorityAdmissionGate`, when $D = 0.0$, `Reserved` requests return `ActionWait` while `Interactive Overflow` requests return `ActionDrop` (HTTP 429).
  * If Worker 1 dequeues a `Reserved` request at the head of the queue, `Apply()` returns `ActionWait`. Worker 1 enters the polling loop inside `Apply()` (`lines 214-233`), waiting for saturation to clear.
  * Because Worker 1 is trapped waiting on that single `Reserved` request and Workers $2 \dots W$ are throttled, **no other worker can dequeue messages from `requestChannel`**.
  * Any `Interactive Overflow` requests waiting behind it—which should be rejected immediately with HTTP 429—are trapped indefinitely in `requestChannel`.

---

### 2. Gate Type Incompatibilities & Semantic Mismatch

#### A. Incompatibility with `LocalConcurrencyGate` in Composite Gates
* **Location:** `LocalConcurrencyGate.Budget(ctx)` (`pkg/async/inference/flowcontrol/local_concurrency_gate.go:52-63`).
* **Why it happens:**
  * `computeActiveWorkers` assumes `Budget(ctx)` represents an **absolute capacity share / target concurrency fraction** ($A = \lfloor W \times D \rfloor$).
  * However, `LocalConcurrencyGate.Budget(ctx)` returns **remaining unused headroom** ($D = (\text{limit} - \text{inFlight}) / \text{limit}$).
* **The Concurrency Collapse Edge Case:**
  * Suppose a pool has $W = 64$ workers and a `LocalConcurrencyGate` with `limit = 50`.
  * As workers pick up requests and `inFlight` reaches 40, `LocalConcurrencyGate.Budget()` returns `(50 - 40) / 50 = 0.20`.
  * Now `computeActiveWorkers(64, 0.20)` evaluates to `12`.
  * Any worker with `workerID > 12` that finishes a request is immediately throttled! Even though 10 concurrency slots are still open (`40 < 50`), the worker pool's active limit collapses to 12 and becomes trapped at low concurrency, **never able to reach `limit = 50`**.

---

### 3. Resource Contention & Observability Spam

#### A. Prometheus Metric Lock Contention & Thundering Herd (1,200+ calls/sec)
* **Location:** `pkg/asyncworker/worker.go:64-78`
  ```go
  for {
      budget := poolGate.Budget(requestCtx)
      metrics.SetDispatchBudget(budget, "", "", poolID)
      ...
      case <-time.After(gateWaitPollInterval): // 50ms
  }

Issue #308, opened 2026-07-13
[Feature]: OpenAI-compatible synchronous gateway in front of the async queues (producer-backed)


shimib:
## Summary

Add a new service that exposes the **OpenAI REST API** (e.g. `/v1/chat/completions`, `/v1/completions`) as a drop-in front end for the async queues. It accepts a normal synchronous OpenAI request, **submits it to the queue via the `producer` library**, waits for the correlated result, and returns it to the caller. The async processor behind the queue applies gates (quota, saturation, priority) and merge policies before forwarding to the inference gateway.

The goal is **frictionless migration**: an operator points their OpenAI client's base URL at this service instead of the llm-d router and transparently gains async's fairness/back-pressure/prioritization — with no client code changes.

## Motivation

Today async is only reachable by writing directly to the queue (via the `producer` library or raw Redis/PubSub). That's a real adoption barrier for teams already speaking the OpenAI API to the llm-d router. This service closes that gap:

- **Migration path:** swap the base URL; keep using the OpenAI SDK.
- **Gain gates + merge policies for free:** per-tenant quota (reserved/overflow), pool saturation back-off, and tier-priority ordering (#294) now apply to real-time traffic, not just batch jobs.
- **Disconnect = cancel:** when a client hangs up, the service can call `CancelRequests` (#299/#307) so abandoned requests are dropped before dispatch instead of wasting inference.

Note it does **not** replace the router — it sits *in front of the queue*; the async processor still forwards to the gateway/EPP/vLLM behind it:

```
OpenAI client → [this service] → queue → async-processor (gates + merge) → inference gateway (EPP) → vLLM
```

## Proposed solution

A new service/binary that:
1. Serves OpenAI-compatible HTTP endpoints.
2. Translates each request into an `api.RequestMessage` (`payload` = the body, `endpoint` = the path, `deadline` = now + request timeout) and submits it via `producer.SubmitRequest`.
3. Blocks until the correlated `api.ResultMessage` arrives, maps it back to an HTTP response, and returns it.
4. On client disconnect / timeout, calls `producer.CancelRequests`.

It should use the `producer` library internally rather than re-implementing queue plumbing.

## Central design challenge — result correlation

`producer.GetResult` is a **shared `BRPOP`** on one result queue (`producer/redis_sortedset_producer.go`), so a single caller draining it receives *all* results, not the one it's waiting for. A synchronous, multi-request, possibly multi-replica front end needs per-request correlation. Options:

[...]

PR #376, opened 2026-07-29
Docs/OpenAI frontend proposal

BenjaminBraunDev:
## What does this PR do?

Adds `docs/openai-frontend-design.md`, the design for an OpenAI API compatible frontend for llm-d-async. Point any OpenAI API compatable client at it and every request gets the async system's tenant, quota, and priority treatment, served one of three ways: enqueue (202 plus fetch by id), wait (held connection), or direct (labeled proxy to the gateway). Default is configurable for requests that doesn't specify, and this doesn't interfere with the existing normal AP queuing functionality.

## Why is this change needed?

Today the only way into the queue is hand-building the internal JSON envelope against the broker, so every adopting team writes a custom producer and result collector. This design gives them a standard OpenAI endpoint instead, namely for existing llm-d-router users to easily incorporate llm-d-async tools into their existing workflow smoothly.

## Related Issues

Related to #308

## Release note
```release-note
NONE
```


jtechapps:
## Current Status

During our previous discussion, we decided that we should implement the `llm-d-async` frontend as a **plugin within `llm-d-coordinator`**. The next steps are to experiment with getting a prototype working with llm-d-coordinator, and then raising proposal to llm-d-router SIG.

---

## Architectural Rationale

| Key Area | Recommendation & Analysis |
| :--- | :--- |
| **Logic & Proxy Reuse** | **`llm-d-coordinator` already implements reverse proxying and core front-end logic.** Importing async-specific logic into the coordinator prevents redundant implementations. |
| **State Management** | **`llm-d-coordinator` remains fully stateless.** Based on recent changes (e.g., Benjamin’s updates), persistent state is offloaded to **Redis**. Direct path handling between the coordinator and async frontend is already nearly identical. |
| **Unified Traffic Routing** | **All request traffic flows through a single endpoint.** Differentiating between real-time and async traffic can be cleanly handled via request headers. |
| **Operational Efficiency** | **Lower maintenance overhead.** Deploying and maintaining two separate services with redundant code paths significantly increases operational complexity. |

---

## Key Technical Decisions

### Direct Mode Handling
* **Retain Direct Mode Support:** Direct mode should be preserved within the coordinator plugin model.
* **Header Behavior:** Supplying the direct mode header allows requests to bypass queueing like standard non-async requests.
* **Quota & Accounting:** Even when bypassing queues, direct mode requests will still accurately update quota and resource tracking.

BenjaminBraunDev:
Closed in favor of #394, This design doc is now outdated and references the closed full frontend PR. Most of the logic is not on the router side so documentation is better placed there.

Issue #369, opened 2026-07-28
The three gate series carry three different `pool_name` values, so none of them join


shimib:
# The three gate series carry three different `pool_name` values, so none of them join

The same running processor, one queue, one pool, emits:

```
llm_d_async_async_gate_metric_value{pool_name="optimized-baseline"}
llm_d_async_async_gate_metric_threshold{pool_name="optimized-baseline"}
llm_d_async_async_dispatch_budget{pool_name="",queue_id="request-sortedset",...}
llm_d_async_async_broker_backlog{pool_name="default",queue_id="request-sortedset",...}
llm_d_async_async_pool_worker_limit{pool_name="default"}
```

Three different values of `pool_name` for one pool. The gate gauges use the gate's own
`poolLabel` (the InferencePool name), while `SetDispatchBudget` derives it locally:

```go
// pkg/redis/sortedset_impl.go:425-429
poolName := ""
if cfg, ok := r.configMap[queueID]; ok {
    poolName = cfg.WorkerPoolID          // empty in a stock single-pool install
}
metrics.SetDispatchBudget(budget, queueID, queueName, poolName)
```

**Impact:** no PromQL can join the gate's decision (`dispatch_budget`) to the inputs that produced
it (`gate_metric_value` / `gate_metric_threshold`) on `pool_name`, which is the obvious way to
build a "why is the gate closed?" panel. You have to know to drop the label and join on
`queue_name`, which the gate gauges do not carry. Cosmetic in a single-pool cluster; it stops
being cosmetic with more than one pool, where the empty label makes `dispatch_budget`
unattributable.

**Suggested fix:** give all five series the pool label from one source — the gate already knows
it. Falling back to `"default"` rather than `""` when `WorkerPoolID` is unset would at least match
the sibling gauges on the same code path.

Evidence: `curl` against the processor's `:9090/metrics`, this run.

---

*Found while standing up an agentic-workload test environment over `llm-d-async` on GKE (llm-d-async `main`, llm-d `main` + PR #2099). Repro scripts and the full findings ledger live outside this repo; happy to attach them. Finding ID `F-018`.*

PR #378, opened 2026-07-30
refactor: consistent config flags for transport implementations

evacchi:
## What does this PR do?

Reorganizes backend/transport configuration around a single **transport** abstraction: each backend is now selected with `--transport` and configured from one JSON document supplied via `--transport-config` (inline) or `--transport-config-file` (file). This replaces the previous sprawl of per-backend CLI flags (`--redis.*`, `--redis.ss.*`, `--pubsub.*`, `--message-queue-impl`).

Key changes:

- **New transport-config surface**
  - `--transport` — `redis-pubsub` | `redis-sortedset` | `gcp-pubsub` (default `redis-pubsub`).
  - `--transport-config` / `--transport-config-file` — inline or file JSON, mutually exclusive, one required when using the new path.
  - Backends parse their config via `redis.LoadPubSubConfig`, `redis.LoadSortedSetConfig`, `pubsub.LoadConfig` (each with `ApplyDefaults`/`Validate`/env overrides), and expose new constructor signatures:
    - `NewRedisMQFlow(cfg, workerPools)`
    - `NewRedisSortedSetFlow(cfg, workerPools, gateFactory)`
    - `NewGCPPubSubMQFlow(cfg, workerPools, gateFactory)`
  - `runner.go`'s `loadFlow` selects by transport type and passes config bytes.
- **Backwards compatibility (`pkg/server/compat.go`)**
  - Every legacy flag stays registered and functional; a `logr` deprecation warning names the replacement when one is used.
  - Legacy flags are translated into the equivalent transport JSON (`synthesizeTransportConfig`), including single-queue fallback and `--redis-tracing` → `enable_tracing`.
  - New flags win when both are set (the legacy ones are then reported as ignored).
  - The retired `gcp-pubsub-gated` implementation is accepted via `--message-queue-impl` and normalized to `gcp-pubsub` with per-topic `gate_type`.
- **Flag naming consistency**
  - Added `--request-merge-policy-config-file` (matching `--transport-config-file`); the older `--request-merge-policy-config` is kept as a deprecated alias. Precedence resolved by `Options.mergePolicyConfigFile()`.
- **Docs**
  - `README.md`: new **Transport Configuration** section with per-backend JSON schemas; per-backend flag sections marked deprecated with alias→field mappings; merge-policy and redis-tracing references updated.

Preserved from main (not regressed): module path `github.com/llm-d/llm-d-async`; Valkey support; EPP budget cascade / gate factory; plugin-based merge policy; redis tracing; backlog polling; and all newer sorted-set/pubsub runtime behavior (cancellation, retry-reservation release #311, gate-closed accounting #368, `GateOwner` stamping #369, structured results, pubsub health probing). Only the config/constructor surface was swapped; runtime logic is main's.

## Why is this change needed?

The per-backend flag surface had grown large and duplicative, with many mutually-exclusive flags per backend. A single JSON config per transport is easier to document, template in Helm, and extend, and it unifies single-queue and multi-queue configuration. All previous flags are retained and still work, so existing deployments keep functioning while migrating.

## How was this tested?

Details:
- `make build` and `make test` pass across all modules (root, `api/`, `pipeline/`, `producer/`); `go fmt`/`go vet` clean.
- Ported the four backend flow tests to the new constructors (keeping main's added tests) and added `pkg/server/compat_test.go` covering config synthesis, new-vs-legacy precedence, gated-alias normalization, deprecation warnings, and the merge-policy resolver.
- `go vet -tags integration ./test/integration/` compiles.
- Manual smoke: `--help` shows both surfaces; validation paths verified for new-path (`--transport-config` required), legacy-path (`--pubsub.project-id` required), and invalid `--transport`.

## Checklist

## Related Issues

Follow up to #283, #287, #295
Related https://github.com/llm-d/llm-d-async/issues/284

## Release note

```release-note
Breaking: Add a unified transport-config surface: select a backend with `--transport`
(`redis-pubsub` | `redis-sortedset` | `gcp-pubsub`) and configure it from one
JSON document supplied via `--transport-config` (inline) or
`--transport-config-file` (file). The previous per-backend flags
(`--message-queue-impl`, `--redis.*`, `--redis.ss.*`, `--pubsub.*`,
`--redis-tracing`) are translated internally and still work, logging a
deprecation warning that names the replacement; `--request-merge-policy-config`
is likewise superseded by `--request-merge-policy-config-file`. **Deprecated:**
these legacy flags will be removed no earlier than `v0.10.0`.

Behavior notes for the legacy path:

- **Breaking (legacy path):** for the Redis transports, `REDIS_URL` is now a
  fallback default for `url` rather than an override — an explicit `--redis.url`
  (or inline `url` in `--transport-config`) wins when both are set. The Helm
  chart, which injects `REDIS_URL` and never sets `url`, is unaffected.
- `--message-queue-impl=gcp-pubsub` (the plain, ungated alias) stays ungated: a
  `gate_type` in its topics-config file remains inert, as before. Per-topic
  gating is available on the new `--transport gcp-pubsub` surface via
  `gate_type`/`gate_params`.
- The `redis-sortedset` single-queue retry fallback now uses the first
  configured queue's `queue_name` (previously `--redis.ss.request-queue-name`).
  In multi-queue configs the fallback is order-dependent on the `queues` array.


Deprecation window: 2 releases from now.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The JSON document is transport-specific. Its `queues`/`topics` entries use the s
 }
 ```
 
-The `url` field honors the `REDIS_URL` environment variable as an override for the Redis transports. Per-queue/topic dispatch gates are configured with the `gate_type`/`gate_params` fields (see [Dispatch Gates](#dispatch-gates)).
+For the Redis transports, `REDIS_URL` is used as a fallback default for the `url` field: an explicit `url` in the transport config takes precedence, and `REDIS_URL` fills it in only when `url` is empty. Per-queue/topic dispatch gates are configured with the `gate_type`/`gate_params` fields (see [Dispatch Gates](#dispatch-gates)).
 
 > **Deprecated:** The per-backend flags — `--message-queue-impl`, `--redis.url`, `--redis.*`, `--redis.ss.*`, `--pubsub.*`, `--redis-tracing`, and `--request-merge-policy-config` — still work for backwards compatibility but are deprecated. When used, the processor logs a warning and translates them into the transport config above. `--transport`/`--transport-config` take precedence when both are set. Prefer the new flags.
 
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ The architecture adheres to the following core principles:
   - [Backend Compatibility](#backend-compatibility)
   - [Deployment](#deployment)
   - [Command line parameters](#command-line-parameters)
+  - [Transport Configuration](#transport-configuration)
   - [Dispatch Gates](#dispatch-gates)
     - [Per-Queue Dispatch Gates](#per-queue-dispatch-gates)
   - [Request Messages and Consumption](#request-messages-and-consumption)
@@ -62,9 +63,9 @@ The architecture adheres to the following core principles:
 
 The Async Processor uses the Redis wire protocol for its message queue implementations (`redis-sortedset`, `redis-pubsub`) and dispatch gates (`redis`, `redis-quota`). Redis-protocol-compatible backends such as [Valkey](https://valkey.io/) can be used with the existing Redis configuration surface.
 
-All `--redis.*` CLI flags, Helm values, and environment variables (e.g. `REDIS_URL`) work unchanged with Valkey — point them at your Valkey endpoint the same way you would with Redis.
+The `url` field in the transport configuration (see [Transport Configuration](#transport-configuration)), the `REDIS_URL` environment variable, and the deprecated `--redis.*` CLI flags all work unchanged with Valkey — point them at your Valkey endpoint the same way you would with Redis.
 
-> **Note:** CLI flags and Helm values retain the `redis.*` prefix because it refers to the wire protocol, not a specific product.
+> **Note:** The `url`/`redis.*` naming is retained because it refers to the wire protocol, not a specific product.
 
 ## Deployment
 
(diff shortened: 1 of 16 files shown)


shimib:
Nice refactor — the compat shim is thorough, tests carried over 1:1 rather than dropped, and adding `--transport-config` to `sensitiveFlags` is a good catch now that credentials live in the JSON. Built and ran unit + integration tests and `golangci-lint` on the branch, all green.

  Two things I'd want addressed before merge, both on the legacy path:

  **1. `REDIS_URL` now beats an explicit `--redis.url`.** `ApplyEnvOverrides()` runs on every load,
  including the shim's synthesized config, and overwrites unconditionally. On main the env only
  seeds the flag's *default*, so the flag wins. Confirmed locally:

  synthesized JSON:  {"url":"redis://from-flag:6379", ...}
  effective URL:      redis://from-env:6379

  The chart is unaffected today (it injects `REDIS_URL` and never passes `--redis.url`), but the
  same inversion means an inline `url` in `--transport-config` is silently ignored whenever `REDIS_URL` is set — which is exactly the chart's environment, so it's a trap for the chart's own  migration to the new surface. Suggest `if c.URL == "" { c.URL = os.Getenv("REDIS_URL") }` so it's  a default rather than an override. README:150 documents it as an override, so if that's deliberate it at least wants a test.

  **2. Legacy `--message-queue-impl=gcp-pubsub` (ungated) now gates.** `loadFlow` passes
  `gateFactory` unconditionally, and `pubsubimpl.go:145` builds a gate when `gateFactory != nil && cfg.GateType != ""`. On main the plain impl passed no factory, so `gate_type` in a topics file was inert and every topic got `ConstOpenGate()`. Same config now starts throttling. (The gated alias itself is fine — its single-topic fallback never set  `gate_type`.) `TestResolveTransport_NormalizesGatedAlias` only asserts the type string, so this isn't covered either way. Either pass a nil factory for the plain legacy impl, or call the change out explicitly in the release note.

  **Missing release-note fragment.** The body has a non-`NONE` release-note block but there's no
  `release-notes.d/unreleased/378.md` — `release-notes-lint` only validates fragments that exist, so
  CI won't catch it. Related: what's the removal window for the legacy flags? Neither the PR nor the
  README commits to a release, and the fragment should say.

  **Chart follow-up (not blocking).** `charts/llm-d-async/templates/ap-deployments.yaml:48+` still
  emits every deprecated flag, so a default install logs ~8 deprecation warnings per pod start about
  flags the operator doesn't control. Worth an issue — `tests/deployment_test.yaml` asserts those
  args, so the two move together.

  Nits:

  - `sortedset_impl.go:137`: `defaultRequestQueueName` moves from `--redis.ss.request-queue-name` to
    `Queues[0].QueueName`. Reasonable fix (the old default rarely matched a real queue in multi-queue
    setups) and it's tested, but it's order-dependent — reordering the array moves the retry
    fallback. Worth a release-note line.
  - `WarnDeprecatedFlags` is exported but only called from `runner.go` in the same package; same for
    `ApplyDefaults`/`ApplyEnvOverrides`/`Validate`, which only their own `LoadX` calls. Per
    CLAUDE.md's YAGNI rule these can be unexported.

evacchi:
> Related: what's the removal window for the legacy flags? Neither the PR nor the
README commits to a release, and the fragment should say.

IIRC the time frame usually is +2 releases 🤔

evacchi:
> Chart follow-up (not blocking).

yes let's create an issue for that

EDIT: PR #388

shimib (approved):

PR #379, opened 2026-07-30
docs(proposal): feedback-based dispatch control for gateway-fronted pools

LukeAVanDrie:
## What does this PR do?

Adds a proposal under `docs/proposals/` for feedback-based dispatch control: sizing dispatch windows from the gateway's own responses instead of scraped metrics.

- A new `aimd` pool gate classifies each response by the router's reason-coded rejection header (`x-llm-d-request-dropped-reason`) and adjusts one window per priority band — additive increase on accepts, multiplicative decrease on capacity rejections, full close for the duration of a `Retry-After`. Evictions and non-capacity errors leave windows unchanged.
- The merge policies stamp the tenant attribute the quota gate already keys on into the router's fairness header (`x-llm-d-inference-fairness-id`), extending tenant isolation past dispatch.
- The proposal records the contract fields with per-field compatibility status (two header names are provisional pending ratification with `llm-d-router`), the control-law design, and the alternatives considered.

Scoped to the async-processor side. Router-side signal work and joint ratification are being worked with the `llm-d-router` maintainers and will be proposed in that repository; nothing here waits on it.

## Why is this change needed?

The current integration couples the processor to the router's internals with no contract underneath: a five-second metric cache against millisecond pool dynamics, capacity constants copied from the router's configuration, one reaction to every 429, and a scrape path that bypasses the gateway. Response feedback replaces the estimation. The signal is as fresh as the traffic it describes, needs no third-party service in the dispatch path, and each contract field the router adds improves the gate without a coordinated release.

## How was this tested?

- [x] Manual testing performed — docs-only PR. The reference implementation, its tests, and an e2e benchmark comparing the gate with the metric-gated configurations follow in separate PRs.

## Checklist

## Related Issues

Related to the AP/flow-control boundary work with the `llm-d-router` maintainers; router-side changes will be proposed in that repository.

## Release note

```release-note
NONE
```

### DIFF
--- a/docs/proposals/feedback-based-dispatch-control.md
+++ b/docs/proposals/feedback-based-dispatch-control.md
@@ -44,13 +44,15 @@ The processor gains one mechanism in each direction of the boundary:
    | Response | Outcome | Window effect |
    |---|---|---|
    | 2xx | accepted | grows (slow start or additive increase) |
-   | reason `rejected-saturated`, or bare 429 | capacity rejection | multiplicative decrease; `Retry-After` closes the window for its duration |
+   | reason `rejected-saturated`, or bare 429 | capacity rejection | multiplicative decrease; `Retry-After` closes the window for its duration, capped by configuration |
    | reason `rejected-ttl-expired` | queue-TTL expiry | gentle decrease (×0.9) |
    | reason `evicted*` | eviction | none |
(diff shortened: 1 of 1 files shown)


jtechapps (approved):

jtechapps:
Tracking issue for the async-processor changes in this feature: #419

PR #380, opened 2026-07-30
feat(mergepolicy): stamp inference fairness ID from tenant metadata

LukeAVanDrie:
## What does this PR do?

Both merge policies (`random-robin`, `tier-priority`) now stamp the tenant attribute the `redis-quota` gate already keys on (default `userid`, from message metadata) into the router's fairness header `x-llm-d-inference-fairness-id`, so the gateway's flow control can arbitrate fairness between tenants after dispatch.

- New policy parameters `fairness_header` / `fairness_attribute`; explicit `fairness_header: ""` disables stamping (a pointer distinguishes absent from empty in config).
- The header is stamped only when the attribute is present, non-empty, and a legal HTTP header value, and a caller-supplied header with the same name under any letter case takes precedence (HTTP header names are case-insensitive; stamping anyway would leave two case variants with nondeterministic wire order).
- Stamping is best effort and never costs a request. Metadata is caller-supplied and never previously reached the wire, so an identity `net/http` refuses to write (a control character) skips the stamp rather than failing the whole dispatch as a non-retryable `UNKNOWN` error.
- The stamping logic lives in one place — the new internal package `pkg/async/mergepolicy/internal/fairness` — instead of being duplicated per policy, and both constructors now take a `Config` struct rather than 4–5 positional strings, so their parameters cannot be transposed and can grow without breaking the signature again.

## Why is this change needed?

Tenant isolation currently ends at dispatch: quotas are enforced upstream, but all of the processor's traffic at a given priority arrives at the gateway as one flow, so one tenant's burst delays every other tenant's requests behind it — the gateway's fairness policies receive no identity to arbitrate on. One header, written by the processor from an attribute both sides already carry, extends isolation past dispatch with no new mechanism on either side. Part of the async-side changes from the feedback-based dispatch control proposal (#379); independently useful and dependency-free.

Deployments whose messages carry no `userid` metadata keep today's behavior exactly. Where the attribute is present, note that stamping is on by default, so its value now reaches the gateway and may be recorded in access logs — the README recommends an opaque tenant ID over personally identifying values, and `fairness_header: ""` opts out entirely.

## How was this tested?

- [x] Unit tests added/updated — stamping behavior is covered once, in the new `internal/fairness` package (stamped identity, configured vs. defaulted attribute, absent and empty attribute, explicit disable, caller precedence including case variants, invalid header value skipped); each policy keeps two tests pinning only that it wires the stamper into its dispatch path and resolves its plugin parameters
- [x] Integration/e2e tests added/updated — existing integration suites updated for the new `Config` constructors and passing
- [x] Manual testing performed — `make ci` green and `go test -race ./pkg/async/mergepolicy/...` clean; verified end-to-end that a `userid` containing CRLF now dispatches successfully with no fairness header stamped, where before the guard it produced a permanent `invalid header field value` failure

## Checklist

- [x] Documentation updated (if applicable) — README merge-policy parameters plus a note on the access-log implication, and the release-note fragment `release-notes.d/unreleased/380.md`

## Related Issues

Related to #379 (feedback-based dispatch control proposal).

## Release note

```release-note
Merge policies stamp the tenant identity attribute (default "userid") into the x-llm-d-inference-fairness-id request header so gateway flow control can arbitrate fairness per tenant; set fairness_header to "" to disable.
```

### DIFF
--- a/pkg/async/mergepolicy/internal/fairness/fairness.go
+++ b/pkg/async/mergepolicy/internal/fairness/fairness.go
@@ -5,7 +5,10 @@
 package fairness
 
 import (
+	"fmt"
 	"strings"
+
+	"golang.org/x/net/http/httpguts"
 
 	"github.com/llm-d/llm-d-async/api"
 )
@@ -14,6 +17,13 @@ import (
 // not configure one. It matches the redis-quota gate's default so both sides key
 // on the same tenant.
 const DefaultAttribute = "userid"
+
+// maxIdentityLen bounds the stamped identity. Envoy's default header limit is
+// around 60 KiB, so an oversized attribute would turn every one of a tenant's
+// requests into a 431 at the gateway, and it amplifies gateway access logs. A
+// tenant ID is far shorter than this in practice; anything longer is skipped
+// rather than stamped.
+const maxIdentityLen = 256
 
 // Params are the fairness parameters every merge policy accepts in its plugin
 // configuration. Embed it in a policy's parameter struct; encoding/json promotes
@@ -25,14 +35,22 @@ type Params struct {
 	Attribute string  `json:"fairness_attribute"`
 }
 
-// HeaderOrDefault returns the configured header name, falling back to
+// ResolveHeader returns the configured header name, falling back to
 // api.FairnessIDHeader when the parameter was absent. An explicit "" is returned
 // unchanged, since that disables stamping.
-func (p Params) HeaderOrDefault() string {
+//
+// It errors when the configured name is not a legal HTTP header name. net/http
+// refuses to write such a header, which would fail every dispatched request
+// permanently as a non-retryable error, so an operator's typo has to surface at
+// startup rather than at dispatch.
+func (p Params) ResolveHeader() (string, error) {
 	if p.Header == nil {
-		return api.FairnessIDHeader
+		return api.FairnessIDHeader, nil
 	}
-	return *p.Header
+	if *p.Header != "" && !httpguts.ValidHeaderFieldName(*p.Header) {
+		return "", fmt.Errorf("fairness_header %q is not a legal HTTP header name", *p.Header)
+	}
+	return *p.Header, nil
 }
 
 // Stamper writes a tenant identity into an HTTP header. The zero Stamper is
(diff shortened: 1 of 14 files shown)


pkg/async/mergepolicy/internal/fairness/fairness.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	headers[s.header] = id
+}
+
+// validHeaderValue reports whether v can be written as an HTTP header value,
+// mirroring what net/http enforces at write time. Metadata is caller-supplied,
+// and net/http fails the entire request on a value it cannot write; a
+// best-effort fairness hint must not cost the request, so an invalid identity
+// skips the stamp instead.
+func validHeaderValue(v string) bool {
```
LukeAVanDrie: We can remove this (or replace it with a dependency) if we feel this guard is overly defensive or brittle.

shimib:
a few comments:

**1. Caller-supplied header wins, so the fairness ID is client-assertable.** `ReqHeaders()` is
  caller-controlled queue data, same trust level as `Metadata`. Since `redis-quota` keys on
  `Metadata[attribute]` (`gate_factory.go:190`), the two can diverge: a tenant can be quota-accounted
  as `A` while presenting fairness ID `B`, or emit a fresh random ID per request so EPP mints a new
  `FlowKey` each time and it escapes fairness sharing entirely. Callers can already override
  `x-gateway-priority` through the same pass-through loop, so this isn't a new class — but `api.go`
  now says stamping means "tenant isolation extends past dispatch", and the code doesn't enforce
  that. Wire-determinism is satisfied equally well by stripping case variants and then stamping.
  Either flip to metadata-wins (or make it configurable), or soften the `api.go`/README wording.

  **2. The configured header *name* isn't validated.** Values are, names aren't.
  `fairness_header: "x-fair id"` (or with `:` / a newline) makes `Do` fail with
  `net/http: invalid header field name`, which maps to `ErrCategoryUnknown` — not retryable
  (`api/inference_error.go:17`) — so *every* request through that policy fails permanently at
  dispatch. Confirmed empirically. That's the same failure mode the value check exists to prevent,
  just on the operator-config side. Validating in the plugin factory would turn it into a startup
  error. (Same gap on `priority_header`, pre-existing.)

  **3. No length bound on the stamped value.** Any length passes `validHeaderValue`; Envoy's default
  ~60 KiB header limit turns a large `userid` into a 431, and it amplifies gateway access logs. A
  256-byte cap with skip-on-exceed would match the existing best-effort posture.

LukeAVanDrie:
Thanks @shimib! All three points are addressed now:

1. Precedence. Flipped to metadata-wins, via your suggestion
2. Header name validation. HeaderOrDefault() became ResolveHeader() (string, error), so both factories reject an illegal name at startup instead of failing every dispatch as non-retryable UNKNOWN. Covers priority_header too.
3. Length bound. 256 bytes, skip-on-exceed, alongside the value check.

shimib (approved):

jtechapps (approved):

BenjaminBraunDev:
Just a heads up: the OpenAI frontend branch (design proposal #376) currently carries a tier-priority commit that also stamps x-llm-d-inference-fairness-id from message metadata, plus a lane_objectives mapping that stamps per-lane InferenceObjective names. Your fairness implementation here is more complete (both policies, header validation, the shared internal package) so my plan is to drop the fairness half from my initial PR and keep only the per-lane objective stamping and then switch to using your config once this merges.

PR #381, opened 2026-07-30
feat(worker): capture gateway drop reasons and honor Retry-After within the deadline

LukeAVanDrie:
## What does this PR do?

Captures the gateway's per-response rejection context on `api.ClientError`:

- `DroppedReason` from the router's `x-llm-d-request-dropped-reason` response header, on every response path — 429, other 4xx, 5xx, and failed body reads (where the headers arrived even if the body did not).
- `Retry-After` on 5xx responses (previously parsed only on 429). It stays scoped to the retryable categories; the fatal 4xx path records the reason only.
- `ClientError.Error()` includes the reason, so it reaches traces (`span.RecordError`) and result payloads today rather than waiting on a consumer.

Downstream policy can then react per reason (capacity vs queue-TTL expiry vs eviction) instead of treating every 429 the same; the first consumer is the feedback-driven dispatch gate proposed in #379.

**Behavior change:** a server `Retry-After` exceeding the message's remaining deadline no longer terminates the message, and hints are no longer honored verbatim. The hint is clamped to half the remaining deadline and jittered up by <25%, with the backoff sample as a floor (shape from @shimib's review: discarding an oversized hint retried fastest when the gateway asked for the longest pause). Previously a 429 with an oversized `Retry-After` returned a deadline-exceeded result with retry budget unused. The server's estimate may be wrong, so it never owns more than half the budget; the message deadline makes the give-up decision. The jitter spreads messages rejected by the same saturation event, which carry identical hints.

That bound makes the second deadline check in `retryMessage` unreachable: the backoff arm jitters strictly below `min(maxDelaySeconds, secondsToDeadline)` and the Retry-After arm is at most `0.625 × secondsToDeadline`. The check is removed in favor of documenting the invariant; deadline expiry is handled at the top of the function, now the single give-up path.

## Why is this change needed?

The router ships machine-readable drop reasons today, and the async processor discards them: the only rejection signal it keeps is the status code. This is the first of the async-side changes from the feedback-based dispatch control proposal (#379); it is independently useful (richer error context, correct `Retry-After` handling on 503s from intermediate proxies) and carries no dependency on the rest.

Worth noting for reviewers: the header is not 429-only. `llm-d-router`'s `translateFlowControlOutcome` sets it on 503 for `rejected-no-endpoints`, `rejected-context-cancelled`, and `rejected-shutting-down`, and on 429 for `rejected-saturated`, `rejected-ttl-expired`, and the `evicted-*` family. That is what motivates capturing on the 5xx path. The [llm-d header reference](https://github.com/llm-d/llm-d/blob/main/docs/api-reference/epp-http-headers.md) still says "Only present on `429`" and omits two reason values — stale relative to the router source; I'll send a separate docs PR.

Naming follows the two existing implementations of this field rather than inventing a third spelling: `RequestDroppedReason` in `llm-d-router` and `DroppedReason` in `llm-d-batch-gateway`. `api/` is a separately tagged module, so the field is public API from the next tag on.

## How was this tested?

  - drop-reason capture across the status/reason pairings the router actually emits, plus a defensive 400
  - the body-read-error path (headers present, body absent)
  - `Retry-After` on 5xx
  - `ClientError.Error()` formatting across all three shapes, including the no-reason case so the existing format stays pinned
  - retry scheduling for `Retry-After` hints: the oversized-hint case (clamped, not terminal), the clamp-binding case (300s hint, 40s deadline → waits 20–25s instead of 2–4s), and jitter bounds on honored hints
- [x] Manual testing performed — `make ci` green; `pkg/asyncworker` coverage 85.8% → 86.6%

Repo hygiene picked up along the way: `make test` did not run the `api` module (`fmt` and `vet` both did), so `api/internal_api_test.go` had never run under the documented test command. Added it — the module now reports 62.2%. `pipeline/` has the same gap but is untouched here and is not covered by `fmt`/`vet` either; that belongs in its own PR.

## Checklist

- [x] Documentation updated (if applicable) — release-note fragment added at `release-notes.d/unreleased/381.md`

## Related Issues

Related to #379 (feedback-based dispatch control proposal).

## Release note

```release-note
The worker now records the gateway's machine-readable drop reason (`x-llm-d-request-dropped-reason`) on rejected requests and surfaces it in error messages, and honors `Retry-After` on 5xx responses as well as 429s. A `Retry-After` hint is now clamped to half the message's remaining deadline instead of failing the message early when it exceeds the deadline, and honored hints are jittered to spread retries from the same rejection event.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,8 +99,9 @@ vet: ## Run go vet against root, api, and producer modules.
 	cd producer && GOWORK=off go vet ./...
 
 .PHONY: test
-test: fmt vet setup-envtest ## Run tests (root module and producer submodule).
+test: fmt vet setup-envtest ## Run tests against root, api, and producer modules.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
+	cd api && go test ./... -coverprofile=cover-api.out
 	cd producer && GOWORK=off go test ./... -coverprofile=cover-producer.out
 
 # Creates a multi-node Kind cluster
--- a/api/inference_error.go
+++ b/api/inference_error.go
@@ -36,6 +36,13 @@ type InferenceError interface {
 
 var _ InferenceError = (*ClientError)(nil)
 
+// DroppedReasonHeader is the response header llm-d-router's flow control uses
+// to communicate the machine-readable reason a request was rejected before
+// dispatch or evicted in flight. The router sets it on both 429 (rejected for
+// capacity or queue TTL, evicted after dispatch) and 503 (no endpoints, client
+// disconnected, shutting down) responses.
+const DroppedReasonHeader = "x-llm-d-request-dropped-reason"
+
 // ClientError represents an inference client error with category and context.
 type ClientError struct {
 	ErrorCategory ErrorCategory
@@ -43,13 +50,18 @@ type ClientError struct {
 	RawError      error         // original error if available
 	RetryAfter    time.Duration // server-specified retry delay from Retry-After header (0 means not set)
 	StatusCode    int           // HTTP status code; 0 means no HTTP response was received
+	DroppedReason string        // machine-readable drop reason from DroppedReasonHeader (empty means not set)
 }
 
 func (e *ClientError) Error() string {
-	if e.RawError != nil {
-		return fmt.Sprintf("%s: %s (caused by: %v)", e.ErrorCategory, e.Message, e.RawError)
+	msg := e.Message
+	if e.DroppedReason != "" {
+		msg = fmt.Sprintf("%s (dropped: %s)", msg, e.DroppedReason)
 	}
-	return fmt.Sprintf("%s: %s", e.ErrorCategory, e.Message)
+	if e.RawError != nil {
+		return fmt.Sprintf("%s: %s (caused by: %v)", e.ErrorCategory, msg, e.RawError)
+	}
+	return fmt.Sprintf("%s: %s", e.ErrorCategory, msg)
 }
 
 func (e *ClientError) Unwrap() error {
--- a/api/inference_error_test.go
+++ b/api/inference_error_test.go
@@ -1,0 +1,46 @@
+package api
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestClientError_Error(t *testing.T) {
+	tests := []struct {
+		name string
+		err  *ClientError
+		want string
+	}{
+		{
+			name: "message only",
+			err:  &ClientError{ErrorCategory: ErrCategoryRateLimit, Message: "rate limited: status code 429"},
+			want: "RATE_LIMIT: rate limited: status code 429",
+		},
+		{
+			name: "with dropped reason",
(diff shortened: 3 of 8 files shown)


shimib:
Suggesting maybe clamping oversized Retry-After rather than discarding.

  if retryAfterSec := retryAfter.Seconds(); retryAfterSec > finalDuration && retryAfterSec < float64(secondsToDeadline) {

  The all-or-nothing condition means a server that asks for a long pause gets retried sooner than one that asks for a short one. Concretely: 429 rejected-saturated, Retry-After: 300, deadline 40s away, RetryCount 0 → the hint is dropped and
  expBackoffDuration(1, 40) schedules a retry in 2–4 seconds. We just told the gateway "you're saturated, here's another one" two seconds after it asked for five minutes.

  I agree with the give-up removal — the deadline should own that decision, not the server. But "ignore the hint" and "terminate the message" aren't the only two options. Clamping keeps both properties:

  if retryAfterSec := retryAfter.Seconds(); retryAfterSec > finalDuration {
      finalDuration = math.Min(retryAfterSec, expBackoffDuration(msg.RetryCount+1, int(secondsToDeadline)))
  }

  or more directly, cap at some fraction just under secondsToDeadline. Either way the message keeps its retry budget and lands before the deadline, but it waits as long as the deadline allows rather than the least amount possible. Retries
  against a saturated gateway aren't free — they consume the flow-control queue slots that are the scarce resource,. Amplifying retry rate precisely when the server signals overload is backwards.

  This only bites at low retry counts (by RetryCount ≥ 5 the backoff is already at the 60s cap), but low retry counts are the common case for a message hitting a saturation event.

LukeAVanDrie:
> Suggesting maybe clamping oversized Retry-After rather than discarding.

Good idea! Reproduced your example: `expBackoffDuration(1, 40)` jitters in [2, 4), so the gateway sees a retry about 2s after asking for 300. Took the clamp, with two changes.

One note on the first snippet: `math.Min(retryAfterSec, expBackoffDuration(...))` re-samples the same distribution, and the min can never exceed that distribution's cap, so an oversized hint is still discarded. The direct form works.

Pushed:

```go
if retryAfterSec := retryAfter.Seconds(); retryAfterSec > finalDuration {
	clamped := math.Min(retryAfterSec, float64(secondsToDeadline)/2)
	// Max keeps the backoff floor when the deadline is nearly spent;
	// jitter de-synchronizes messages rejected by the same event.
	finalDuration = math.Max(finalDuration, clamped*(1+rand.Float64()/4))
}
```

- Clamped at half the remaining deadline. Clamping to nearly all of it would let one wrong server estimate consume the whole budget in a single wait. Half caps the cost at half the budget before jitter (~62.5% after) and leaves the rest to the backoff ladder.
- Jittered the honored hint. Messages rejected by one saturation event carry the same Retry-After and would re-land together if honored verbatim.

Bounds: clamped ≤ deadline/2, jittered < 0.625×deadline, and the backoff arm is already strictly under the deadline, so every retry lands before expiry. The `math.Max` matters when the deadline is nearly spent: deadline/2 can fall below the backoff sample there, and a bare clamp would retry sooner than the current code.

This also clamps hints between deadline/2 and deadline that the previous head honored verbatim. That is a second behavior change; the PR description covers both. Your 300s/40s case now waits [20, 25).

shimib (approved):

PR #382, opened 2026-07-30
feat(flowcontrol): aimd dispatch gate driven by response feedback

LukeAVanDrie:
## What does this PR do?

Adds an `aimd` pool dispatch gate that sizes per-priority-band dispatch windows from gateway response feedback instead of scraped metrics — the core mechanism of the feedback-based dispatch control proposal (#379).

- One congestion window per `(classification, tier)` band: slow start below `ssthresh`, additive increase on accepts, multiplicative decrease on capacity rejections (coalesced per flight of sends via a TCP-style recovery point), full close for a server `Retry-After` with slow-start reopen. Advisory views refine the controller when present: advertised band headroom caps the window; queue duration past a target holds or gently shrinks it.
- When a band is full or closed, only reserved requests park the worker (`Wait`); overflow is refused back to the broker, so sheddable backlog never captures pool workers, and its redeliveries double as the band's probe traffic.
- Bands are coupled the way the gateway's strict-priority dispatch couples them: a rejection decreases its band and every band below, holds growth above; an accept clears the holds above it; a queue-TTL expiry gently shrinks its band and propagates the same hold. Band order and naming come from a new shared `api.LaneRank`/`api.LaneLabel`, the single source of the lane mapping the `tier-priority` merge policy also uses.
- Composition with the existing gate vocabulary: `composite` and `wait-on-refuse` forward feedback and wait signals to inner gates; `tier-priority-admission` reads any non-Continue inner verdict as saturation; queue-level consumers degrade reserved `Wait` verdicts to re-enqueue.
- Observability: `async_aimd_window` / `_ssthresh` / `_inflight` gauges labeled `(pool_name, band)`.

The worker-side wiring that feeds the gate real response outcomes follows in a separate PR; until it lands, the gate admits against its windows but observes no outcomes.

## Why is this change needed?

The metric gates estimate pool capacity from outside: cached Prometheus reads, a capacity formula with constants copied from the router's configuration, and no per-reason rejection handling. The proposal (#379) documents the failure modes (staleness, drift, conflated rejections, bypass). Response feedback is exactly as fresh as the traffic it describes, needs no third-party service in the dispatch path, and works against an unmodified `llm-d-router` today. The metric gates remain the supported path for backends with no gateway.

## How was this tested?

- [x] Unit tests added/updated — controller behavior (window limits and wake, slow start/congestion avoidance, per-flight decrease coalescing, Retry-After close and reopen, headroom cap, queue-duration signal, TTL handling, ladder coupling in both directions, stale-evidence guards), reserved-park/overflow-refuse verdicts, wrapper forwarding and fan-in, tier-priority-admission with a waiting inner gate, factory parameter validation, `api.LaneRank`/`api.LaneLabel` tables
- [ ] Integration/e2e tests added/updated — a Kind clearing-time benchmark comparing this gate against the metric-gated pools follows with the worker wiring
- [x] Manual testing performed — `make ci` green

## Checklist

- [x] Documentation updated (if applicable) — `docs/aimd-gate.md`, README gate list and parameters

## Related Issues

Implements the gate proposed in #379. The two advisory view header names remain provisional pending ratification with `llm-d-router`.

## Release note

```release-note
Added an "aimd" pool dispatch gate that sizes per-priority-band dispatch windows from gateway response feedback (reason-coded rejections, Retry-After, advisory capacity views), removing the Prometheus dependency for gateway-fronted pools.
```


LukeAVanDrie:
This lands at ~1.4k lines, but ~560 of them are tests and ~95 docs. If it's easier to review in pieces, I'm happy to split it into three PRs, in dependency order:

1. **Lane mapping**: `api.LaneRank`/`api.LaneLabel` plus the `tier-priority` merge policy refactored onto them, and the `api` module tests wired into `make test`. No behavior change, ~120 lines.
2. **Feedback contract**: the `pipeline` types (`DispatchFeedback`, `FeedbackGate`, `WaitNotifier`), wrapper forwarding in `composite`/`wait-on-refuse`, and the queue-level `Wait` degradation. Inert until a feedback gate consumes it, ~250 lines.
3. **The `aimd` gate**: controller, factory registration, gauges, docs, and the test suite. ~1k lines, dominated by the gate/test file pair.

If reviewing it whole, the read order I'd suggest: `docs/aimd-gate.md` → `pipeline/feedback.go` → `aimd_gate.go` → the tests (the design rationale is in #379, not here).

shimib:
1. PoolLabel is read from user config, bypassing GateOwner.

  gate_factory.go does PoolLabel: paramString(params, "pool", ""). pipeline/gate.go documents GateOwner as "stamped by the message queue implementation at construction time and is never read from user config, so a gate's own metrics can carry
  the same queue_id/queue_name/pool_name labels as the owner's other series" — and pkg/server/runner.go:146 already populates Owner: pipeline.GateOwner{WorkerPoolID: poolID} for pool gates. So the pool param is both redundant and divergent: it
  can disagree with the pool_name on every other series for the same pool, and its default of "" means two aimd pool gates collide on the same (pool_name="", band) gauge series and clobber each other.

  Should be PoolLabel: cfg.Owner.WorkerPoolID, with the pool param and its README/doc entries dropped.

  2. ObserveOutcome with a nil Msg silently attributes to overflow/batch and holds every band.

  band(nil) → tier "" → batch, classification overflow → rank 5. gate_forwarding_test.go:82 and :104 call ObserveOutcome with a nil Msg and assert only "no panic" — but the real behavior is that it creates and mutates the lowest band. A nil-Msg
  OutcomeRejectedCapacity then holds all five bands above it. Since the producer doesn't exist yet, this is the moment to pin the contract: either document Msg as required on DispatchFeedback and drop feedback with a nil Msg on the floor, or state explicitly that unattributable feedback lands in overflow/batch. Right now the tests encode the accident as the contract.

  3. WaitSignal() can return nil while the type assertion always succeeds. CompositeGate and WaitOnRefuseGate both declare var _ pipeline.WaitNotifier, so gate.(pipeline.WaitNotifier) is always ok — but both can return nil, and
  TestCompositeGate_NoNotifiersMeansNilSignal asserts exactly that. The obvious consumer shape (if n, ok := g.(WaitNotifier); ok { select { case <-n.WaitSignal(): ... } }) then blocks forever on a nil channel. The WaitNotifier doc in
  feedback.go:72-79 doesn't mention it. Either document "may return nil, treat as no wake support" on the interface, or have the wrappers not implement the interface when they have nothing to forward (harder, given they're constructed before that's known — so documenting is fine, just do it on the interface where the consumer will read it).

  4. One wake per event, shared across bands. notify is buffered 1 with a non-blocking send, so when a window grows by 4 or several slots free together, exactly one parked worker wakes and the rest ride the poll timer. Cross-band too: a freed
  overflow/batch slot can consume the token and wake a worker parked on reserved/interactive, which re-applies, fails, re-parks. The polling fallback makes this a latency issue rather than a stall, and the "spurious wake" comment acknowledges it — but the docs claim parked workers are "woken as soon as a slot frees or the window grows," which overstates it. Either widen to a per-band signal or soften the doc.

  5. NewCompositeGate fan-in goroutines never exit. for range signal over a channel that's never closed, one goroutine per inner notifier, with the comment "live as long as the process." True for production (composites are built once at startup) but not for tests, which construct them freely. If this ever grows a dynamic reconfiguration path it becomes a real leak. A context or stop chan on the composite would cost little now.

LukeAVanDrie:
@shimib, thanks for the feedback; all should be addressed in the latest commit

1. PoolLabel: now stamped from `GateConfig.Owner.WorkerPoolID`; the `pool` param is gone from the factory, README, and docs.
2. nil-Msg feedback: dropped in `ObserveOutcome` rather than attributed to the lowest band; `DispatchFeedback.Msg` is documented as required, and `TestAIMDGate_NilMsgFeedbackIsDropped` pins it (replacing the tests that encoded the accident).
3. nil WaitSignal: documented on the `WaitNotifier` interface: may be nil, a nil channel blocks forever, treat nil as poll-only.
4. Shared wake token: took the "soften the doc" option: interface doc, `wake()` comment, README, and docs/aimd-gate.md now describe the wake as a shared hint with a poll fallback. A per-band signal needs the real park loop to shape it, so I'd rather design that in the worker-wiring PR.
5. Fan-in goroutines: `CompositeGate.Close()` (idempotent stop channel) ends them; the fan-in test uses `t.Cleanup`. The fan-in also now skips inner gates whose `WaitSignal()` is nil.

LukeAVanDrie:
/hold

@jtechapps and @shimib

Followup from our discussion earlier:

---

**TODO before this merges (or as an immediate follow-up): cap the Retry-After close.**

`ObserveOutcome` currently trusts `Retry-After` verbatim: `closedUntil = now.Add(fb.RetryAfter)`, with the close inherited by every band below. The proposal (#379) now specifies the close as a bounded hint ("capped by configuration"), so the gate needs to match:

- Add a config field (e.g. `max_retry_after`) and clamp the close duration where `closedUntil` is set in the `OutcomeRejectedCapacity` branch. The inherited close on lower bands is copied from `closedUntil`, so the clamp covers it automatically.
- Reasoning, recorded in the proposal: the header is advisory (RFC 9110), the router's projection will ride on a drain-rate estimate that starts crude, and a closed band produces no responses that could correct a wrong projection. Unbounded trust turns one bad `Retry-After: 3600` into an hour-long, ladder-wide outage. A cap bounds the cost per event.
- Default TBD: a small multiple of `hold_duration` keeps it in the same time scale as the other cooldown, or a flat few seconds. Decide when implementing.
- This bound is deliberately different from the worker-side one being discussed on #381 (clamp to a fraction of the message's remaining deadline): a message has a deadline budget, a band only has config. Same bounded-hint principle with different denominators.
- Optional, second-order: jitter the close duration. All inherited closes and all replicas that saw the same event reopen at the same instant. Unlike the per-message herd on #381, reopen here starts at MinWindow in slow start, so the burst is one probe per band per replica.

This is dead code today (nothing emits the header), which is exactly why it should land before router-side emission does. Optional extra, can wait: keep one probe slot while closed instead of going fully deaf, so a wrong projection gets refuted by evidence rather than by the timer.

jtechapps:
Unstale

PR #383, opened 2026-07-31
feat(chart): support pod affinity (ap.affinity) for the llm-d-async chart

todayim:
## Summary

Add optional `ap.affinity` to the `async-processor` Helm chart so the dispatcher Deployment's pod spec can carry affinity rules.

## Motivation

In production clusters the Async Processor (dispatcher for async batch inference) typically belongs on CPU/infra nodes, while GPU nodes are either tainted or too valuable to host a small control-plane-ish workload. Without chart-level affinity support, operators have to post-patch the rendered Deployment (or fork the chart) just to keep the pod off GPU pools. A first-class `affinity` value keeps such clusters on the stock chart.

## Changes

- `templates/ap-deployments.yaml`: render `.Values.ap.affinity` into the Deployment's pod spec (`affinity` omitted entirely when unset — behavior is byte-identical for existing values)
- `values.yaml`: document `ap.affinity` (default `{}`) with a nodeAffinity example

## Testing

- `helm template` renders the affinity block correctly when set, and is unchanged when unset
- Deployed on Kubernetes 1.26 serving 3-model mixed vLLM/SGLang async batch traffic via `redis-sortedset` queues; dispatcher pinned to infra nodes throughout a 900-request mixed load run

## Checklist

- [x] Signed-off-by included (DCO)
- [x] No behavior change for existing releases (empty default + `with` guard)

### DIFF
--- a/charts/llm-d-async/templates/ap-deployments.yaml
+++ b/charts/llm-d-async/templates/ap-deployments.yaml
@@ -241,3 +241,7 @@ spec:
             secretName: {{ .Values.ap.tls.secretName }}
         {{- end }}
       {{- end }}
+      {{- with .Values.ap.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
--- a/charts/llm-d-async/tests/deployment_test.yaml
+++ b/charts/llm-d-async/tests/deployment_test.yaml
@@ -417,3 +417,23 @@ tests:
       - notContains:
           path: spec.template.spec.containers[0].args
           content: --transform-config-file=/etc/llm-d-async/config/transform-config.json
+
+  - it: should not render affinity by default
+    asserts:
+      - isNull:
+          path: spec.template.spec.affinity
+
+  - it: should render affinity when set
+    set:
+      ap.affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: node-role.kubernetes.io/infra
+                    operator: In
+                    values: ["true"]
+    asserts:
+      - equal:
+          path: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].key
+          value: node-role.kubernetes.io/infra
--- a/charts/llm-d-async/tests/deployment_test.yaml
+++ b/charts/llm-d-async/tests/deployment_test.yaml
@@ -431,9 +431,13 @@ tests:
             nodeSelectorTerms:
               - matchExpressions:
                   - key: node-role.kubernetes.io/infra
-                    operator: In
-                    values: ["true"]
+                    operator: Exists
     asserts:
       - equal:
           path: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].key
           value: node-role.kubernetes.io/infra
+      - equal:
+          path: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].operator
+          value: Exists
+      - notExists:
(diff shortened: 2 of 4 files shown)


shimib:
Verified locally rather than eyeballing: with `ap.affinity` unset, `helm template` output is
  byte-identical to `main`, so the no-behavior-change claim holds. Set, it renders correctly under the
  pod spec. `helm lint` clean, `helm unittest` 54/54 pass. Single Deployment with no `range`, so the
  root-context lookup is right, and the `{{- with }}` + `toYaml | nindent 8` shape matches how the
  chart already handles `ap.securityContext` and `ap.resources`. Nice, tight change.

  One ask before merge:

  **Add helm-unittest coverage.** `tests/deployment_test.yaml` consistently gives every optional value
  both a set and an unset case (TLS volume, `transformConfig`, resources), and CI gates on
  `helm unittest`. I ran this against your branch and it passes (56/56):

  ```yaml
    - it: should not render affinity by default
      asserts:
        - isNull:
            path: spec.template.spec.affinity

    - it: should render affinity when set
      set:
        ap.affinity:
          nodeAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
              nodeSelectorTerms:
                - matchExpressions:
                    - key: node-role.kubernetes.io/infra
                      operator: In
                      values: ["true"]
      asserts:
        - equal:
            path: spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].key
            value: node-role.kubernetes.io/infra
```
  Minor:
  - No release-notes.d/unreleased/383.md. Precedent is mixed for chart-only PRs (#359/#360 shipped
  without), but this adds a user-facing value and release-notes-lint won't catch the absence.
  - The title says async-processor; the chart was renamed to llm-d-async in v0.8.0 (#340). Paths in
  the diff are correct — just worth fixing the title, since it becomes the squash-commit subject.
  - Placement between terminationGracePeriodSeconds and volumes splits the volumes conditional
  from its matching volumeMounts block above. Purely cosmetic, take it or leave it.
  - The chart has no nodeSelector/tolerations/topologySpreadConstraints today. Affinity alone is
  genuinely sufficient for your case (tainted GPU nodes repel by default), so I wouldn't widen this
  PR — but worth filing a follow-up if you want the rest.

todayim:
Thanks for the careful review! All addressed in c74639f:

- **helm-unittest coverage added**: both cases (unset → `affinity` is null; set → renders under pod spec) appended to `tests/deployment_test.yaml`. Verified locally: `helm unittest charts/llm-d-async` 56/56 pass, `helm lint` clean.
- **Release note**: added `release-notes.d/unreleased/383.md` following the front-matter precedent.
- **Title**: fixed to reference the llm-d-async chart.
- **Placement**: good catch — moved the affinity block below the `volumes` block so the volumeMounts/volumes pairing stays together.

Agreed on not widening scope here; will file a follow-up for nodeSelector/tolerations/TSC if needed.

shimib:
Hi @todayim, thanks for adding the test coverage and release note fragment so quickly!

  Now that I'm looking at the concrete unit test and values example, I have two technical points regarding scheduling correctness in real-world clusters, plus two quick cosmetic adjustments:

  1. Valueless label convention in the nodeAffinity example
  In Kubernetes, node-role.kubernetes.io/* labels are valueless by convention—an infra node is labeled node-role.kubernetes.io/infra: "", not "true".

  The documented example uses:

   1 - key: node-role.kubernetes.io/infra
   2   operator: In
   3   values: ["true"]
  Because of this, an operator who copy-pastes this example into their values will match zero nodes, and the dispatcher pod will be stuck in Pending forever.

  To fix this, we should change the example to use the Exists operator (and drop the values array):

   1 - key: node-role.kubernetes.io/infra
   2   operator: Exists
  Since the new unit test in tests/deployment_test.yaml mirrors the exact In expression from the example, we should update both the values.yaml comments and the test case to use the proper Exists shape.

 Thank you again for driving this feature forward!

todayim:
Addressed in 10a7bc7:

- Updated the values.yaml nodeAffinity example to use `operator: Exists` for the valueless `node-role.kubernetes.io/infra` label, without `values`.
- Updated the helm unittest fixture the same way and added assertions for `operator: Exists` and no `values` field.

Verified locally: `helm unittest charts/llm-d-async` (56/56 pass), `helm lint charts/llm-d-async` clean.

shimib (approved):

PR #385, opened 2026-07-31
Add async_gate_decisions_total metrics to worker pool gates

### LINKED ISSUE #384: [Metrics] Add gate decision metric for worker pool gates
gate metrics such as async_gate_decisions_total are not configured for worker pool gate decisions.

jtechapps:
## What does this PR do?

Previously, async_gate_decisions_total (via metrics.RecordGateDecision)
and series initialization (via metrics.InitGateDecisions) were only
included in queue-level gates (Pub/Sub and Redis sorted set consumers).
Worker pool-level gates did not record gate decision reasons or pre-create
zero-valued series in Prometheus.
This change:
* pkg/asyncworker/worker.go:
  - Calls metrics.RecordGateDecision(...) across all worker pool gate
    verdicts (error, dropped, gate_closed, and quota_exhausted for
    ClassificationOverflow messages).
  - Hoists a waitRecorded boolean outside the gating loop so requests parked
    in ActionWait record gate_closed exactly once per message instead of on
    every 50ms poll tick.
* pkg/server/runner.go:
  - Calls metrics.InitGateDecisions("", "", poolID) during pool gate
    construction (if pool.GateType != "") so gated worker pools pre-create
    all four reason series at 0 on startup without polluting gateless pools.
* README.md:
  - Updates the llm_d_async_async_gate_decisions_total table description
    and pool-level paragraph to document that both pool-level gate gauges
    and async_gate_decisions_total are keyed by pool_name alone with empty
    queue_id/queue_name labels.
* pkg/asyncworker/worker_test.go:
  - Adds TestWorker_PoolGateDecisionsMetrics with subtests for dropped,
    gate_closed, quota_exhausted, error, and wait_decision_-_counts_once_per_message.
* release-notes.d/unreleased/add-pool-gate-decisions-metrics.md:
  - Adds release note fragment for PR #385.

Intentionally did not include `async_dispatch_budget` metric because `gate.Budget()` has not effect for worker pool gates. This will only be applicable after https://github.com/llm-d/llm-d-async/issues/304, https://github.com/llm-d/llm-d-async/issues/141

## Why is this change needed?

Fixes: #384 

## How was this tested?

## Checklist

## Release note

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,7 +626,7 @@ The Async Processor exposes Prometheus metrics under the `llm_d_async` subsystem
 | `llm_d_async_async_queue_residence_time_millis` | Histogram | Time in milliseconds a message spent buffered in-process, from broker ingestion until a worker pulled it for processing. Measures the async delay introduced by the system (queue time). Always registered. |
 | `llm_d_async_async_dispatch_budget` | Gauge | Current dispatch budget [0.0–1.0] returned by the queue's gate; the fraction of system capacity available for new requests (0.0 = gate fully closed). Useful for diagnosing why throughput is throttled. |
 | `llm_d_async_async_pool_worker_limit` | Gauge | Configured worker concurrency limit for a pool (carries only the `pool_name` label). Compare against `llm_d_async_async_inflight_requests` to compute worker utilization. |
-| `llm_d_async_async_gate_decisions_total` | Counter | Count of gate decisions that prevented dispatch, by `reason`: `gate_closed` (no dispatch budget), `quota_exhausted` (per-attribute quota overflow), `dropped` (gate permanently rejected the request), `error` (gate evaluation failed). `quota_exhausted`, `dropped` and `error` count individual messages refused after being dequeued. `gate_closed` counts those plus every dequeue round in which the budget shrank the batch to zero — the way budget-based gates (`prometheus-budget`/`-saturation`/`-query`) shed work *before* a message is dequeued — so its rate reflects throttled dispatch opportunities, not messages. All four `reason` series are created at 0 when a queue starts, so a query returns 0 rather than an empty vector. |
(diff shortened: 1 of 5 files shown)


shimib:
Nice catch on the gap — pool gates being invisible in `async_gate_decisions_total` is a real hole, and the label choice `("", "", poolID)` is right: it matches the pool-gate convention already documented in the README ("no single queue, so … keyed by `pool_name` alone") and the `GateOwner{WorkerPoolID: poolID}` in `runner.go`. One issue in the `ActionWait` branch though.

  **`ActionWait` counts once per poll tick, not once per message.**

  `worker.go:186` sits inside the `for {}` retry loop, so `RecordGateDecision` fires on every iteration while a worker is parked. `gateWaitPollInterval` is 50ms, so a parked request adds ~20 counts/sec for as long as it waits. I ran a gate that always returns `Wait()` with a 2s deadline against a single request:

  ```
  ONE request parked ~2s -> gate_closed count = 40
  ```

  This isn't an edge case — `wait-on-refuse` converts every `Refuse` into `Wait` by design, and `tier-priority-admission` returns `Wait` for every saturated reserved request. Those are the two gates most likely to be on a pool.

  It also contradicts the metric's documented semantics in both the `Help` string and the README: *"`quota_exhausted`, `dropped` and `error` count individual messages refused after being dequeued."* And it inflates the README's own example query  — `sum by (pool_name) (rate(...{reason="gate_closed"}[5m]))` becomes a function of the poll interval rather than the workload, making a `wait-on-refuse` pool and a plain-`Refuse` pool incomparable on the same panel.

  Suggest hoisting a `waitRecorded bool` outside the loop and recording once on entry to the wait — the message is the unit of accounting, and it stays parked either way.

  The new test covers `dropped` / `gate_closed` / `quota_exhausted` / `error` but not `ActionWait`, which is where this lives. Worth a fifth subtest asserting the count is exactly 1 after parking through several poll intervals.

  Smaller things:

  - `runner.go:238` inits the series for *every* pool, but the gate is only built later under `if pool.GateType != ""`. Gateless pools get four permanently-zero series. Fine if intentional — maybe move the init next to the gate construction, or note the intent in a comment.
  - The README metric table spells out exactly which decisions are counted; this widens that scope, and the pool-level paragraph currently only mentions gauges. Worth a sentence.
  - Release note is `NONE`, but a new series under an existing metric is user-facing — on upgrade, queue-level and pool-level decisions for the same pool both contribute (from distinct label sets), so existing `sum by (pool_name)` panels will shift. I'd add a fragment.
  - Nit: the `msg.InternalRequest != nil` guard is inconsistent with `poolGate.Apply(gateCtx, msg.InternalRequest, ...)` one line above, which is unguarded — and `EmbelishedRequestMessage`'s own doc says the embedded request "must be non-nil in normal use". The queue-level sites call `ir.GetClassification()` unguarded; I'd match them.

jtechapps:
Updated PR to address comments.

shimib (approved):

PR #388, opened 2026-08-03
refactor: update helm config to new format

### LINKED ISSUE #284: [Feature]: unify config across backends

evacchi:
## What does this PR do?

Updates helm charts and docs to follow the config changes in #378. Stacked, pending merge there, and also cross-testing on llm-d-batch-gateway. Relevant commit is just 2b579b0

## Why is this change needed?

#378 harmonizes configuration style for llm-d-async, this propagates the change to the Helm charts. The old Helm format still works, and also the old flags work for the time being. In v+2 we will remove the deprecated formats.

## How was this tested?

## Checklist

## Related Issues

Follow up to #378, closes https://github.com/llm-d/llm-d-async/issues/284

## Release note

```release-note
Breaking: The Helm chart now configures the processor through the unified transport
surface. Select the backend with `ap.transport`
(`redis-pubsub` | `redis-sortedset` | `gcp-pubsub`) and configure it with a
single `ap.transportConfig` document (mirroring `--transport` /
`--transport-config`). Redis/Valkey credentials still come from `ap.redis.url`
or `ap.redis.secretName` (injected as `REDIS_URL`); leave `url` out of
`ap.transportConfig` to use them.

The deprecated per-backend values (`ap.messageQueueImpl`, the `ap.redis.*` /
`ap.gcpPubSub.*` request/queue fields, and `ap.otel.redisTracing`) still work —
the chart translates them into the transport config — and the install notes
print a migration warning when they are used. Either way the chart now emits
only `--transport` / `--transport-config`, so a default install no longer logs
per-flag deprecation warnings from the processor.

Deprecation Window: 2 versions from now.
```

### DIFF
--- a/charts/llm-d-async/templates/NOTES.txt
+++ b/charts/llm-d-async/templates/NOTES.txt
@@ -1,0 +1,24 @@
+llm-d-async {{ .Chart.AppVersion }} has been installed as {{ include "llm-d-async.fullname" . }}.
+
+Transport: {{ include "llm-d-async.transport" . }}
+
+The processor is configured through the unified transport surface
+(--transport / --transport-config); no deprecated CLI flags are emitted.
+{{- if include "llm-d-async.usingDeprecatedTransport" . }}
+
+WARNING: deprecated chart values in use.
+  You configured the transport with the deprecated per-backend values
+  (ap.redis.enabled / ap.gcpPubSub.enabled / ap.messageQueueImpl and the
+  per-backend ap.redis.* / ap.gcpPubSub.* request/queue fields, ap.otel.redisTracing).
+  They still work — the chart translates them into the transport config — but
+  will be removed in a future release. Migrate to the unified surface:
+
+    ap:
+      transport: {{ include "llm-d-async.transport" . }}
+      transportConfig:
+        # transport-specific JSON document; see the chart README and the
+        # "Transport Configuration" section of the project README.
+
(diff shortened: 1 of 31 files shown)


jtechapps (dismissed):

evacchi:
rebased/force-pushed

jtechapps (approved):

PR #392, opened 2026-08-04
feat(frontend): OpenAI-compatible frontend with enqueue, wait, and direct modes

### LINKED ISSUE #308: [Feature]: OpenAI-compatible synchronous gateway in front of the async queues (producer-backed)

BenjaminBraunDev:
#### What does this PR do?

Adds an optional OpenAI API compatible HTTP frontend as a new top-level module, as well as some small AP-side changes for it to function.

The frontend serves three modes selected per request by header with a configurable default:
- enqueue (202 plus fetch by id)
- wait (held connection with keyspace-notification wake-up and a capped fallback to the enqueue response)
- direct (labeled reverse proxy to the inference gateway, with streaming)

Results land in tenant-scoped keys with a TTL and are reclaimed after delivery. Direct mode classifies against the same Redis quota counters the AP's gates use. AP-side changes: result_ttl_seconds on queue config, DEADLINE_EXCEEDED classification for deadline-aborted sends, and per-lane objective stamping in the merge policy.

#### Why?

Implements the design in `docs/openai-frontend-design.md` (#376). Stock OpenAI clients can use llm-d-async without hand-building the broker envelope or draining the shared result list.

#### How was this tested?

Unit tests for the frontend module (83% coverage) and the AP changes, make ci across all four modules, and E2E on an H200 GKE cluster: all three modes, the labeling chain to the EPP, streaming passthrough, deadline aborts on both paths, tenant-scoped fetch denial, a replica kill under 200 held wait connections with zero client errors, Redis restart fail open, and load tests from 400 QPS direct through GPU saturation with clean 429 and 504 degradation. Confirmed fairness and objective stamping on both the queued and direct paths.

#### How to review

Diff anatomy: 84% of the lines are self-contained frontend/module code (its own Go module, ~1200 lines of implementation, ~550 of tests, plus generated go.sum and a README). Another 6 percent is the chart template and Dockerfile, inert unless frontend.enabled is set. Changes to existing AP code is only 79 lines across four files (result TTL, deadline classification, lane objectives, Makefile), each in its own commit with tests. Reviewing commit by commit follows that split cleanly.

#### Related Issues

Fixes #308
Design doc #376

#### Release note
```release-note
Add an optional OpenAI API compatible frontend serving enqueue, wait, and direct modes against existing async queues.
```


BenjaminBraunDev:
Closed in favor of #394

PR #394, opened 2026-08-07
feat(ap): result TTLs, per-lane objective stamping, and deadline classification

BenjaminBraunDev:
#### What does this PR do?

Adds three AP-side changes that let llm-d-routers `async-broker` plugin serve requests through the broker:

- `result_ttl_seconds` on queue config: the result writer applies an expiry to the result destination on push, so per-request result keys clean themselves up when never fetched
- Per-lane objective and fairness stamping in the tier-priority merge policy, so every dispatch carries InferenceObjective and fairness headers the EPP understands
- DEADLINE_EXCEEDED classification wherever a deadline kills a request

And queues without the new settings behave exactly as today.

#### Why?

These are the broker-side prerequisites: fetchable per-request results need TTL cleanup, dispatches need priority headers the gateway understands, and deadline expiry needs to be distinguishable from generic failure. The serving door itself lives in llm-d-router as a coordinator pipeline step (async-broker), so this is the llm-d-async side of that design.

#### How was this tested?

Unit tests in each commit, make ci across all four modules, and end to end on a GKE cluster behind llm-d-router: TTLs observed on result keys, objective and fairness headers confirmed in EPP logs at dispatch, and mid-flight expiries surfacing as DEADLINE_EXCEEDED.

#### How to review

Non-test code is about 105 lines across three files (pkg/redis, pkg/async/mergepolicy/tierpriority, pkg/asyncworker).

#### Related Issues

Part of #308.

This PR carries the AP-side of the former implementation #392, which proposed a dedicated standalone frontend. The coordinator-based approach serves the same modes through the llm-d-router pipeline instead (llm-d/llm-d-router/pull/2325), so this is the alternative.

#### Release note
```release-note
Support result_ttl_seconds on queue config, stamp per-lane objective and fairness headers at dispatch, and classify deadline-aborted sends as DEADLINE_EXCEEDED.
```

### DIFF
--- a/pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
+++ b/pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
@@ -19,15 +19,15 @@ import (
 func init() {
 	plugins.MustRegister("tier-priority", func(name string, parameters json.RawMessage, handle plugins.Handle) (plugins.Plugin, error) {
 		var params struct {
-			PriorityHeader string `json:"priority_header"`
-			TierLabel      string `json:"tier_label"`
+			PriorityHeader  string `json:"priority_header"`
+			TierLabel       string `json:"tier_label"`
+			ObjectiveHeader string `json:"objective_header"`
 			// LaneObjectives maps lane keys ("reserved-interactive",
 			// "overflow-batch", ...) to InferenceObjective names stamped as
 			// ObjectiveHeader, replacing the per-queue objective for lane
 			// granularity. The numeric priority header has no consumer in
 			// upstream gateways. Objectives are how priority reaches them.
-			ObjectiveHeader string            `json:"objective_header"`
-			LaneObjectives  map[string]string `json:"lane_objectives"`
+			LaneObjectives map[string]string `json:"lane_objectives"`
 			fairness.Params
 		}
 		if len(parameters) > 0 {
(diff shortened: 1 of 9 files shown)


pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
```diff
@@ -21,6 +21,13 @@ func init() {
 		var params struct {
 			PriorityHeader string `json:"priority_header"`
 			TierLabel      string `json:"tier_label"`
+			// LaneObjectives maps lane keys ("reserved-interactive",
```
jtechapps, line 25: nit: move this below ObjectiveHeader

pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
```diff
@@ -36,13 +43,21 @@ func init() {
 		if !httpguts.ValidHeaderFieldName(params.PriorityHeader) {
 			return nil, fmt.Errorf("invalid tier-priority parameters: priority_header %q is not a legal HTTP header name", params.PriorityHeader)
 		}
+		if params.ObjectiveHeader == "" {
+			params.ObjectiveHeader = "x-llm-d-inference-objective"
```
jtechapps: Add "x-llm-d-inference-objective" as const to api/api.go

pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
jtechapps, line 375: This is overridden by the lane objectives. We should add a note that the lane objectives overrides the objective set by the queue.

pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
BenjaminBraunDev: done

pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
BenjaminBraunDev: done

pkg/async/mergepolicy/tierpriority/tier_priority_policy.go
BenjaminBraunDev: added a comment here

jtechapps:
Also forgot to mention. Update the README to add the new tier-priority merge policy changes.

BenjaminBraunDev:
Done. One extra change alongside this: messages the expire in the queue now write DEADLINE_EXCEEDED result instead of being dropped at pop so that deadline cancels report at both queue and mid-flight.

jtechapps (approved):

Issue #1, opened 2026-01-15
[Tracking] Async Processor tasks


shimib:
### Tasks and Roadmap

#### Core
- [ ] Add More Metrics
- [ ] Polish Logs
- [ ] Add End-to-end tests
- [ ] Add a Circuit-breaker

#### Features
- [ ] Fixed-rate backoff support
- [ ] Traffic throttling (by saturation metrics)
- [ ] Request-level SLO

#### Message Queue Implementations

- [ ] GCP PubSub Implementation
- [ ] Kafka support
- [ ] Redis/Valkey Streams support

shimib:
Superseded by the v0.9.0 roadmap tracker #248 (milestone: https://github.com/llm-d-incubation/llm-d-async/milestone/1). Closing this stale tracking issue in favor of #248.

PR #395, opened 2026-08-10
Add GCP PubSub e2e tests using emulator

### LINKED ISSUE #83: Add PubSub queue impl to e2e tests
### Problem Statement

Currently only Redis is being tested as a Queue implementation in the e2e tests

### Proposed Solution

Use GCP PubSub emulator for testing as well

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

jtechapps:
## What does this PR do?

Add GCP PubSub transport testing to the end-to-end test suite using the GCP PubSub emulator.
- Add Kubernetes manifest for PubSub emulator (gcr.io/google.com/cloudsdktool/google-cloud-cli:emulators) exposed via NodePort.
- Update Helm chart deployment template (ap-deployments.yaml) to support injecting custom environment variables via `ap.env`.
- Add Helm configuration (pubsub.yaml) for gcp-pubsub transport configured with PUBSUB_EMULATOR_HOST.
- Add PubSub test utility functions (enqueuePubSubMessage, enqueuePubSubMessages, popPubSubResult) to utils_test.go.
- Add e2e_pubsub_test.go covering end-to-end request processing, batching, and 5xx retries.

## Why is this change needed?

Fixes #83 

## How was this tested?

## Checklist

## Release note

```release-note
NONE
```

### DIFF
--- a/charts/llm-d-async/templates/ap-deployments.yaml
+++ b/charts/llm-d-async/templates/ap-deployments.yaml
@@ -131,6 +131,9 @@ spec:
                 name: {{ include "llm-d-async.redisSecretName" . }}
                 key: {{ include "llm-d-async.redisSecretKey" . }}
           {{- end }}
+          {{- with .Values.ap.env }}
+          {{- toYaml . | nindent 10 }}
+          {{- end }}
         name: llm-d-async
         ports:
           - name: metrics
--- a/charts/llm-d-async/tests/deployment_test.yaml
+++ b/charts/llm-d-async/tests/deployment_test.yaml
@@ -458,3 +458,23 @@ tests:
       - notContains:
           path: spec.template.spec.containers[0].args
           content: --request-merge-policy-config=/etc/llm-d-async/config/request-merge-policy.json
+
+  - it: should render extra env vars when ap.env is set
+    set:
+      ap.env:
+        - name: PUBSUB_EMULATOR_HOST
+          value: "pubsub-emulator:8085"
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: PUBSUB_EMULATOR_HOST
+            value: "pubsub-emulator:8085"
+
+  - it: should not render extra env vars when ap.env is empty
+    asserts:
+      - notContains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: PUBSUB_EMULATOR_HOST
+            value: "pubsub-emulator:8085"
--- a/charts/llm-d-async/values.yaml
+++ b/charts/llm-d-async/values.yaml
@@ -51,6 +51,12 @@ ap:
   drainTimeout: "2m"
   prometheusURL: ""
   prometheusCacheTTL: ""
+  # Extra environment variables to inject into the processor container.
+  # Useful for passing backend-specific configuration like PUBSUB_EMULATOR_HOST.
+  # env:
+  #   - name: PUBSUB_EMULATOR_HOST
+  #     value: "pubsub-emulator:8085"
+  env: []
   # Named worker pools configuration. When specified, overrides the global concurrency option.
   # workerPools:
   #   - id: "default"
--- a/charts/llm-d-async/values.yaml
+++ b/charts/llm-d-async/values.yaml
@@ -53,6 +53,8 @@ ap:
   prometheusCacheTTL: ""
   # Extra environment variables to inject into the processor container.
   # Useful for passing backend-specific configuration like PUBSUB_EMULATOR_HOST.
+  # Note: Avoid setting NAMESPACE or POD_NAME here as it will override the downward
+  # API and break downstream $(NAMESPACE) / $(POD_NAME) OTel resource expansion.
   # env:
   #   - name: PUBSUB_EMULATOR_HOST
   #     value: "pubsub-emulator:8085"
--- a/release-notes.d/unreleased/395.md
+++ b/release-notes.d/unreleased/395.md
@@ -1,0 +1,7 @@
+---
+pr: 395
+url: https://github.com/llm-d/llm-d-async/pull/395
+author: jtechapps
+date: 2026-08-18
+---
+The `llm-d-async` Helm chart now supports `ap.env`, allowing operators to inject arbitrary environment variables into the async-processor container (such as `PUBSUB_EMULATOR_HOST` for local emulation or custom provider configurations).
--- a/test/e2e/e2e_benchmark_test.go
+++ b/test/e2e/e2e_benchmark_test.go
@@ -7,12 +7,14 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
 
+	"cloud.google.com/go/pubsub/v2"
 	"github.com/llm-d/llm-d-async/api"
 )
 
@@ -56,6 +58,7 @@ var _ = ginkgo.Describe("Async Processor Performance Benchmark E2E", ginkgo.Orde
 		rdb.Del(ctx, benchmarkResultQueue)          //nolint:errcheck
 		rdb.Del(ctx, benchmarkPoolGateRequestQueue) //nolint:errcheck
 		rdb.Del(ctx, benchmarkPoolGateResultQueue)  //nolint:errcheck
+		recreatePubSubSubscription(ctx, pubsubClient, pubsubProjectID, pubsubBenchResultSub, pubsubBenchResultTopic)
 	})
 
 	ginkgo.It("measure clearing time and saturation under load with queue-level gate", func() {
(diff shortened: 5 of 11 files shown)


jtechapps:
## Review — GCP PubSub e2e tests

Overall this is a great addition that closes a real coverage gap, and the infra wiring is well done (emulator reachable from both the in-cluster processor and the test process against the same instance; `PUBSUB_EMULATOR_HOST` correctly routes the SDK to the emulator; defensive startup retries; uses the new `transport`/`transportConfig` surface). A few issues below, one of which is a blocker.

### 1. `popPubSubResult` can ack-and-drop results → batch test will flake/fail (blocker)
`client.Subscriber(subName)` uses default `ReceiveSettings` (MaxOutstandingMessages ~1000, NumGoroutines > 1), so `Receive` can deliver **several messages concurrently**. The callback acks *every* delivered message but only stores the first:

```go
if result == nil { ... result = &r }
msg.Ack()   // acks even when result was already set
cancel()
```

In the batch test the processor publishes 3 results close together (batch_size 10). The first `popPubSubResult` grabs all 3, captures one, **acks and discards the other two**. The next loop iterations then find an empty subscription → `Eventually` times out → failure. The single-message and retry specs happen to be safe (only one result in flight), so this only bites the batch case.

Suggested fix — serialize delivery and don't ack what you don't keep:

```go
sub := client.Subscriber(subName)
sub.ReceiveSettings.MaxOutstandingMessages = 1
sub.ReceiveSettings.NumGoroutines = 1
...
if result == nil {
    var r api.ResultMessage
    if json.Unmarshal(msg.Data, &r) == nil { result = &r; msg.Ack() } else { msg.Nack() }
    cancel()
} else {
    msg.Nack()   // leave it on the sub for the next pop
}
```

### 2. No drain of the result subscription between specs (medium)
Results accumulate on the shared result sub with no cleanup between specs. A straggler from the batch test could make the retry test's `Consistently(...ShouldBe Nil())` fail, or a stale ID get popped by the wrong spec. Consider draining the result sub in `BeforeEach` (loop `popPubSubResult` until nil) — composes well with fix #1.

### 3. Unused constants + duplicated literals (medium — may fail `make lint`)
`pubsubRequestSub` and `pubsubResultTopic` (utils_test.go) appear unused — the setup builds full paths as string literals (e.g. `"projects/test-project/subscriptions/pubsub-e2e-request-sub"`) instead of using them. staticcheck (U1000) flags unused unexported consts, so this may break the linter gate. Deriving the resource paths from the constants (e.g. `fmt.Sprintf("projects/%s/topics/%s", project, pubsubRequestTopic)`) removes both the duplication and the dead consts.

### 4. Floating emulator image tag (low–medium)
`gcr.io/google.com/cloudsdktool/google-cloud-cli:emulators` is a mutable tag, unlike the other pinned e2e images (`valkey:8-alpine`, `epp:v1.5.0`, `sim:v0.10.0`). Pin to a concrete version for reproducible runs.

### 5. `ap.env` is undocumented and untested (low)
The chart change is clean and backward-compatible (`{{- with .Values.ap.env }}`, correct `nindent 10`), but it's not added to `values.yaml` with a comment/default, and there's no helm-unittest case despite the chart's strong unit-test suite. A small test asserting `ap.env` entries render into the container `env:` (and nothing renders when unset) would lock it in.

### 6. Minor
- `_ = sub.Receive(...)` swallows all errors; deadline-exceeded is expected, but a not-found/permission error would vanish silently. Consider logging non-deadline errors.
- Inconsistent `AlreadyExists` handling: the request topic retries inside `Eventually`, but the result topic/subs use `Expect(err).NotTo(HaveOccurred())`. Fine functionally, just uneven.

### Security / conventions / coverage
- No security concerns — the emulator is unauthenticated by design, `ap.env` is standard chart flexibility, no secrets involved.
- Follows existing suite patterns well (env-configurable ports/images, `kubectlApplyFile` templating, redeploy lists, AfterSuite close).
- The three scenarios are the right ones; the main gap is harness correctness (#1), not scenario coverage.

**Verdict:** #1 is a blocker (batch spec will be flaky/failing as written); I'd also fix the likely lint failure (#3) before CI, then #2/#4 for stability/reproducibility. #5/#6 are polish.

shimib:
Some comments:


1. Result bleed between specs (Flake Risk)
  popPubSubResult returns any result, and GCP Pub/Sub is at-least-once. The batch spec calls cancel() inside the callback, meaning buffered-but-undispatched messages are returned to the subscription unacked.

  drainPubSubResults cannot reliably clean these up between specs because it breaks on the first 3s empty read and cannot see messages whose ack deadlines (typically 10s) haven't expired yet. When the subsequent retry spec executes:

   1 gomega.Consistently(...).Should(gomega.BeNil())
  The ack deadline expires, the leftovers are redelivered, and the test suddenly flakes. (The Redis suite dodges this because getResultCount uses a non-destructive LLEN).

   * Suggested Fix: Recreate the result subscription inside BeforeEach of the test suite. Since it's running against a local emulator, recreating the subscription is extremely cheap, clean, and guarantees perfect isolation.

  ---

  2. Untested Pub/Sub Gating Path
  In pubsubimpl.go:512-526, the Pub/Sub-specific gating logic scales MaxOutstandingMessages based on the budget and parks the receive loop (ReasonGateClosed) when budget ≤ 0.

  However, test/e2e/helm/pubsub.yaml configures no gate_type under topics, meaning the budget remains pinned at 1.0 and this crucial block is never executed. Let's make sure we track a follow-up issue to add a gated Pub/Sub E2E test, rather
  than closing out the epic on transport plumbing alone.

  ---

  3. Consistently samples ~2×, not 5×
  popPubSubResult uses a hard timeout of 3 seconds:

   1 cctx, cancel := context.WithTimeout(ctx, 3*time.Second)
  Because Gomega evaluates the callback synchronously, a 3s block plus the 1s poll interval means a 5-second Consistently window is only sampled twice instead of five times.
   * Suggested Fix: Drop the pop timeout to ~1s or widen the Consistently duration to ensure a robust sampling run.

  ---

  4. Dead branch & untested attribute-merge path
  In utils_test.go, we have:

   1 if msg.Metadata != nil { pubMsg.Attributes = msg.Metadata }
  This never fires because makeRequestMessage never populates Metadata. Consequently, pubsubimpl.go:576-581 (which merges attributes into request metadata) is completely untested and dead code.
   * Suggested Fix: Add a test spec that publishes attributes and asserts they arrive as metadata (aligning with CLAUDE.md YAGNI guidelines).

  ---

  Nits & Polish

   * Gomega Consistently lockup: In popPubSubResult, if a message fails to unmarshal, it currently calls msg.Nack(). This triggers an immediate redelivery loop, trapping the test thread in an infinite CPU-burning cycle. Calling msg.Ack() and
     logging the error is safer in a test helper.
   * Redundant Mutex: The mu mutex in popPubSubResult is redundant because MaxOutstandingMessages and NumGoroutines are both explicitly set to 1.
   * OTel Environment Collision: ap.env appends directly at the bottom of the container's env block. We should add a quick comment in values.yaml warning operators that manually setting NAMESPACE or POD_NAME in ap.env will silently break the
     downstream $(NAMESPACE) OTel resource expansion.
   * Duplication: os.Setenv("PUBSUB_EMULATOR_HOST", ...) is called twice in test/e2e/e2e_suite_test.go (on lines 479 and 716). We can easily clean this up into a single set at suite initialization.
   * Semantic wording: "collects a batch of messages" tests that three messages are successfully processed individually rather than batching. Re-wording the test description to "processes multiple messages" matches the coverage better.

  Outstanding work getting the Pub/Sub emulator up and running

jtechapps:
I also added queue clearing benchmark test for PubSub transport.

### E2E Flow Control Benchmark Results

We added a performance benchmark spec testing the **GCP Pub/Sub** transport with a Prometheus saturation dispatch gate (`prometheus-saturation` on `e2e-pool` with a `0.8` threshold) under simulated load (5,000 requests, 10ms TTFT, 1ms ITL, max concurrency 32).

#### Benchmark Comparison (5,000 Requests)

| Scenario | Transport | Batch Size | Requests | Queue Clearing Time | Avg Saturation | Status |
| :--- | :--- | :---: | :---: | :---: | :---: | :---: |
| **Queue-Level Gate** | `redis-sortedset` | 64 | 5,000 | **~50–57s** | 0.4320 | **PASS** |
| **Worker Pool Gate** | `redis-sortedset` | 64 | 5,000 | **~55s** | 0.7556 | **PASS** |
| **Pub/Sub Transport Gate** | `gcp-pubsub` | 64 | 5,000 | **3m 16s** | 0.2959 | **PASS** |
| **Pub/Sub Transport Gate** | `gcp-pubsub` | 256 | 5,000 | **1m 34s – 2m 20s** | 0.5829 | **PASS** |

---

### Why is GCP Pub/Sub Slower Than Redis Queue-Level Gating?

Under active flow control and backend saturation, GCP Pub/Sub exhibits slower queue clearing times than Redis due to the following architectural and implementation factors:

#### 1. Gate Budget Recovery Cadence (50ms vs 10s)
* **Redis Sorted Set**: Operates via an active pull loop governed by `poll_interval_ms: 50`. As soon as backend saturation falls below threshold, Redis detects it and resumes full-rate dispatch within **50 milliseconds**.
* **GCP Pub/Sub**: Uses a streaming pull subscriber (`sub.Receive`). In `pkg/pubsub/pubsubimpl.go`, the subscriber dynamically resizes `sub.ReceiveSettings.MaxOutstandingMessages = int(batchSize * budget)`. When the gate closes (`budget <= 0`), the worker waits on a **10-second ticker** (`time.NewTicker(10 * time.Second)`) before checking if the budget has recovered to restart consumption. Every gate closure incurs up to a 10s idle penalty.

#### 2. Nack / Refuse Backoff Penalty (10s Delay)
* In Pub/Sub, messages already pulled into the SDK receive buffer right before saturation spikes are evaluated by `gate.Apply`. If refused, `pkg/pubsub/pubsubimpl.go` delays **10 seconds** before issuing a `Nack()` (`const quotaExceededNackDelay = 10 * time.Second`) to avoid tight spin-loops against the broker.
* In Redis, unpopped or refused items remain in the sorted set and are immediately re-evaluated on the next 50ms tick.

#### 3. Budget Discretization (`math.Floor(batch_size * budget)`)
* Pub/Sub calculates outstanding message quota as $\lfloor \text{batchsize} \times \text{budget} \rfloor$.
* With a smaller batch size (e.g., 64), any fractional budget below $1/64 \approx 0.0156$ truncates to 0, pausing message consumption. Raising `batch_size: 256` gives more headroom during partial recoveries and cut the clearing time by more than half.

shimib:
Two things about the benchmark:

  It can't fail for gate reasons. The only assertion is receivedCount == numRequests. NewSaturationDispatchGate uses fallback = 1.0 - 0.0 = 1.0, so if Prometheus is unreachable or the series is missing, the gate returns full budget, every
  queryProm returns NaN and is filtered out, and the spec prints Average Saturation Level: 0.0000 over an empty slice and passes. One line fixes it:

  gomega.Expect(saturationMetrics).ToNot(gomega.BeEmpty())

  (Same gap exists in the pool-gate benchmark, so no objection to deferring — but worth fixing in both.)

  batch_size: 256 with concurrency: 64 makes the throttle mostly inert. Max budget is 0.8, so MaxOutstandingMessages tops out at 204 while only 64 workers can dispatch. The gate only becomes the binding constraint below budget < 64/256 = 0.25, i.e. saturation above ~0.55. batch_size: 64 would make the scaling observable across the whole range.

  The attributes spec doesn't test attributes. preserves caller attributes and metadata via GCP PubSub sets msg.Metadata and then asserts only result.ID — same coverage as the basic spec. It can't do better as written, since
  api.ResultMessage.Metadata is json:"-" and never round-trips. Either observe it where it's actually visible, or rename to match what it checks (e.g. "accepts messages carrying caller attributes") so the name doesn't overstate.

  Release note. ap.env is a user-facing chart addition, but the PR is marked release-note: NONE with no release-notes.d/unreleased/395.md.

  Nits:
  - recreatePubSubSubscription discards the delete error; that's right for not-found, but any other failure resurfaces as a confusing AlreadyExists from the create. Checking codes.NotFound and asserting otherwise keeps it legible.
  - Three 5000-request benchmarks now, each with a 10-minute Eventually. Last nightly was 850s against a 45m job timeout — margin is thinner than it looks.

jtechapps:
Updated the benchmark to set the batch_size back to 64. 

### Performance Benchmark Results (5,000 Requests under Simulated Load)

| Scenario | Transport | Concurrency | Batch Size | Gate Type | Queue Clearing Time | Avg Saturation | Gate Metric Points Collected | Status |
| :--- | :--- | :---: | :---: | :--- | :---: | :---: | :---: | :---: |
| **Queue-Level Gate** | `redis-sortedset` | 64 | 64 | `composite` (`prometheus-saturation` + `local-max-concurrency`) | **~50s – 1m 00s** | 0.4320 – 0.4400 | 25–30 points (spikes to 1.6) | **PASS** |
| **Worker Pool Gate** | `redis-sortedset` | 64 | 64 | `composite` (`prometheus-saturation` + `local-max-concurrency`) | **~55s** | 0.6519 – 0.7556 | 27 points (spikes to 3.6) | **PASS** |
| **Pub/Sub Transport Gate** | `gcp-pubsub` | 64 | 64 | `prometheus-saturation` (threshold 0.8) | **~3m 16s** | 0.2959 | 98 points (spikes to 3.4) | **PASS** |

### Ket Takeaways

1. Gate Flow-Control Verification:
    * All scenarios actively assert non-empty Prometheus saturation metric collections (gomega.Expect(saturationMetrics).ToNot(gomega.BeEmpty())), proving the dispatch gates were actively throttled during backend saturation.

1. Matching Concurrency & Batch Size (64):
    * Setting batch_size: 64 on Pub/Sub matches the 64 workers, ensuring MaxOutstandingMessages = math.Floor(64 * budget) actively constraints concurrency across the entire saturation spectrum.

1. Queue Clearing Timing:
    * Redis recovers from gate closure in 50ms (poll_interval_ms: 50), clearing in ~50–60s.
    * Pub/Sub recovers via a 10s subscriber budget ticker and a 10s delayed Nack backoff, clearing in ~3m 16s without message loss or overwhelming the backend.

shimib (approved):

Issue #377, opened 2026-07-30
Typos detected by nightly scan (2 found)


clubanderson:
## Typos Detected

The nightly typo checker found **2** typos.

```
./pkg/async/inference/flowcontrol/gate_factory.go:318: `mis` -> `miss`
./release-notes.d/unreleased/374.md:10: `mis` -> `miss`
```

### Fix
```bash
cargo install typos-cli  # or: brew install typos-cli
typos --write-changes .
```

Add false positives to `_typos.toml`.

---
*Nightly scan from [llm-d-infra](https://github.com/llm-d/llm-d-infra/actions/runs/30515318895)*

clubanderson:
## New typos found (2026-08-08)

```
./pkg/redis/sortedset_impl.go:594: `Ded` -> `Dead`
```

Detected by [nightly scan](https://github.com/llm-d/llm-d-infra/actions/runs/31239251049)

clubanderson:
## New typos found (2026-08-09)

```
./pkg/redis/sortedset_impl.go:594: `Ded` -> `Dead`
```

Detected by [nightly scan](https://github.com/llm-d/llm-d-infra/actions/runs/31294525444)

clubanderson:
## New typos found (2026-08-10)

```
./pkg/redis/sortedset_impl.go:594: `Ded` -> `Dead`
```

Detected by [nightly scan](https://github.com/llm-d/llm-d-infra/actions/runs/31355615125)

clubanderson:
## New typos found (2026-08-11)

```
./pkg/redis/sortedset_impl.go:594: `Ded` -> `Dead`
```

Detected by [nightly scan](https://github.com/llm-d/llm-d-infra/actions/runs/31458494760)

clubanderson:
## New typos found (2026-08-12)

```
./pkg/redis/sortedset_impl.go:594: `Ded` -> `Dead`
```

Detected by [nightly scan](https://github.com/llm-d/llm-d-infra/actions/runs/31563507725)

PR #398, opened 2026-08-11
Fix/typo fix

alvin123:
## What does this PR do?

This Cahnge adds two ignore RE rules to skip typos check

## Why is this change needed?

The issue discribed in issue #377 should not be a error since mis-sized makes sense over miss-sized. To address this issue, added ignore RE rule to skip similar issue so that any word starting with mis- is treated as meaningful and valid.
Similar for ZADDed which is also reqported as error by the typos tool, but it apparantly is a valid function name for this product. 

## How was this tested?
Run the 

- [x ] Manual testing performed 

## Checklist

- [x ] Code follows project [contributing guidelines](CONTRIBUTING.md)
- [ x] Tests pass locally (`make test`)
- [ x] Linters pass (`make lint`)

## Release note

```release-note
NONE
```

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -7,7 +7,6 @@ IST = "IST"
 # `unmarshaling` is the standard Go/American spelling (typos suggests British).
 typ = "typ"
 unmarshaling = "unmarshaling"
-mis-sized="mis-sized"
 
 [files]
 extend-exclude = [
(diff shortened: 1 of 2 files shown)


thisnamewasnottaken:
There's already a PR for this ticket.

PR #400, opened 2026-08-15
feat(metrics): deadline proximity histogram for backlog queues.

### LINKED ISSUE #345: [Metrics] Deadline Proximity histogram (time-to-deadline of queued items)
Split out of #217 (comprehensive metrics), which is now closed with the core suite delivered. This is a post-v0.9.0 fast-follow.

### What
A histogram of `deadline - now()` across items currently in the queue, to spot when a large batch of work is about to expire simultaneously (SLO early warning).

- Metric: `async_deadline_proximity_millis` (histogram), labeled by pool/queue.

### Approach
Piggyback the existing per-queue backlog poller (the `ZCard` loop + ticker that powers `async_broker_backlog`): add a bucketed `ZRANGEBYSCORE`/`ZCOUNT` read of the SortedSet scores (deadlines) and observe `deadline - now()` per sampled item.

### Caveats
- **redis-sortedset only** — Pub/Sub can't peek at un-consumed messages, so this metric is not feasible for the redis-pubsub broker (document the limitation).
- Sampled on the poller interval (a snapshot per tick), not continuous.

0xA8hinav:
## What does this PR do?

Implements the deadline proximity histogram (`llm_d_async_async_deadline_proximity_millis`) from #345, as an SLO early warning for batches of queued work about to expire simultaneously.

- `QueueBacklog` (`redis-sortedset`) now reads exact cumulative counts in one `TxPipelined` per queue: `ZCard` + `ZCOUNT` per bucket boundary (expired + 13 buckets, up to 24h). No member payloads are ever read, so cost per poll is independent of prompt size.
- `QueueBacklogStat` gains `ExpiringCounts`, `nil` when the broker cannot report per-item deadlines (Cloud Pub/Sub).
- `pollBacklog` feeds the snapshot histogram; on a failed read the snapshot is zeroed so the metric clears instead of going stale, while `Depth` is preserved so `async_broker_backlog` doesn't look drained.

## Why is this change needed?

Feature request #345: today there is no way to tell that a large batch of queued items is about to hit its SLO deadline simultaneously — items can expire as `DEADLINE_EXCEEDED` in bulk with no advance warning. The histogram shows the time-to-deadline distribution of queued items as a per-poll snapshot (`le="0"` holds items past their deadline but still queued).

## How was this tested?

- [x] Unit tests added/updated (metrics, pipeline, sortedset, poller)
- [x] Manual testing performed — live against Redis 7: queue frozen via the dispatch gate; 5 items at `+120`/`+120`/`+600`/`+600`/`+3600s` produced `le=120000=2`, `le=600000=4`, `le=3600000=5`, `_count=5` (cumulative nesting correct), and `le=0=2` two minutes later after the `+120s` items expired. Screenshot in comments.

## Screenshots

Screenshot_2026-08-21_02-03-00

## Checklist

- [x] Commits are signed off (`git commit -s`) per DCO (`PR_SIGNOFF.md`)
- [x] Code follows project contributing guidelines (`CONTRIBUTING.md`)

## Related Issues

Fixes #345

## Release note

```release-note
Added deadline proximity histogram (`llm_d_async_async_deadline_proximity_millis`) as an SLO early warning for batches of queued work approaching deadlines.

### CODE REVIEW

**Comment by 0xA8hinav** [comment]
@jtechapps 
Merge conflict and ci issue resolved, can you run the ci again?

**Comment by jtechapps** [comment]
I will test this tomorrow. `ZRANGEBYSCORE` performance is the main concern noted below.

 ## AI detected Issues & suggestions

  1. Performance — ZRANGEBYSCORE ships full member payloads (medium, worth discussing)
  The members in this sorted set are the full serialized InternalRequest JSON (the integration test confirms Member: string(irBytes) with the prompt payload). To
  compute proximity you only need the scores, but ZRANGEBYSCORE WITHSCORES transfers every member body too. At the 1000 cap, on a default ~2s poll, that's up to 1000
  full request payloads per queue per tick pulled over the wire and immediately discarded — potentially MB/s per queue for large prompts. The code comment
  acknowledges "there is no scores-only read," which is true for the range command, but a small EVAL Lua script iterating the range and returning only scores would
  eliminate the payload transfer entirely. Given the author already notes the histogram is the less useful of the two metrics (gauge preferred for alerting), it's
  worth weighing the histogram's cost against its value.

  2. metrics.ProximityWindows() allocates 3× per queue per poll (minor perf)
  It's a defensive copy (append([]ProximityWindow(nil), …)), and in the QueueBacklog hot loop it's called three times per queue — for the countCmds capacity hint, in
  the command-building loop, and again in the result-building loop. Call it once and reuse the local:
  windows := metrics.ProximityWindows()

  3. Gauge staleness on transient secondary failure (minor, note the tradeoff)
  On secondaryErr, the views go nil and pollBacklog emits nothing — but async_queue_expiring_items is a gauge, so it retains its last value until the next successful
  poll. This is inconsistent with the deliberate "report 0 for Depth on failure to avoid staleness" choice one branch up. For Redis (always deadline-capable) this
  only affects transient blips and is arguably fine, but it's worth a one-line comment acknowledging that expiring-items can be briefly stale after a failed read,
  since the surrounding code goes to lengths to avoid exactly that for Depth.

  4. Poller tests rely on go … + time.Sleep (minor test-flakiness risk)
  TestPollBacklogObservesDeadlineViews starts the poller and sleeps 50ms against a 10ms interval, asserting >= 2 samples. The tolerance is generous so it'll almost
  always pass, but sleep-based timing tests can flake under CI load. Not blocking.

  5. Style nit
  Building the score bounds mixes strconv.FormatInt(now, 10) and fmt.Sprintf("(%d", now) in adjacent lines — harmless, just slightly inconsistent.

## Performance notes

### The problem, precisely

The histogram only needs each item's score (the deadline). But the current read:

```
  rangeCmd = pipe.ZRangeByScoreWithScores(ctx, cd.queueName, &redis.ZRangeBy{
      Min: "-inf", Max: "+inf", Offset: 0, Count: proximitySampleSize,
  })
```

  ZRANGEBYSCORE ... WITHSCORES returns member and score for every element. In this sorted set the member is the full serialized InternalRequest JSON (prompt payload
  included — the integration test seeds Member: string(irBytes)). So up to 1000 full request bodies are serialized onto the socket, sent over the network, and
  deserialized into []redis.Z on the Go side — every poll tick — only for the scores to be read and the members thrown away.

jtechapps:
The suggestion of using EVAL to strip the members on the server side seems incorrect as Redis will still need to read the members and will block when that happens.

The inference requests can be quite large 100k input tokens may not be uncommon. 100Kb*4*1000 = 400Mb of data processed each tick.

jtechapps:
Actually do we even need both metrics? or `ZRANGEBYSCORE` since histogram can be built from `ZCOUNT`

0xA8hinav:
@jtechapps 
Thank you for the follow-up. I agree with your point on `EVAL`, and I think it also settles the bigger question: the per-item read is too costly, so I want to move the histogram to bucket-based `ZCOUNT`.

**On `EVAL`**
I checked how Redis implements sorted set commands. `ZRANGEBYSCORE`, with or without a script, walks the sorted set and builds the full member strings on the server side. That is where the cost comes from: the members are the whole request payloads, so reading 1,000 of them means the server touches up to **1,000 × prompt size** per queue per tick, and the script only throws the strings away after. 

There is no command that returns scores without members (`ZSCORE`, `ZMSCORE`, and `ZRANK` all take a member as input), so any per-item read is proportional to the payload size no matter how it is done. The commands that only compare scores, like `ZCOUNT` and `ZCARD`, never touch the member bytes.

So I do not think we should keep `ZRANGEBYSCORE` for the histogram. At 100k input tokens per request, the 400MB per tick number you mentioned is real, and the cap at 1,000 only limits *how many* we read, not *how big* each read is.

**The Bucket-Based Approach**
* **`ZCOUNT` per bucket boundary (expired + 13 buckets):** The same pipelined read we already use for the expiring gauge. It never reads payloads, so the cost per tick is ~300 bytes instead of up to 400MB. Exact counts, no sampling, no quantile bias.
* **Optional `ZRANGEBYSCORE LIMIT 0 1`:** For the exact soonest expiry (about one payload per tick). I would keep this read optional and off by default.

**Metric Consolidation**
The expiring-items gauge is already `ZCOUNT`-based. Its windows line up with the histogram bucket boundaries, so the bucket histogram would be the same read with 14 ranges instead of 6 windows. The gauge then steps aside - the histogram replaces it (one metric instead of two) with the same signal it already had.

**The Tradeoff**
We lose exact per-item values, and because Prometheus histograms only accept `Observe()`, we would need a small custom collector with an estimated `_sum`. I think that is a small price to pay for removing the payload cost, and it matches the SLO early warning goal from the issue. 

If you see a reason the per-item values are needed, please tell me 
otherwise personally I would suggest we should go with the bucket-based histogram.

jtechapps:
@Abhinav-kodes LGTM on the bucket based approach.

0xA8hinav:
@jtechapps 
I have pushed the bucket-based implementation.

### What changed
- The histogram now comes from 14 cumulative `ZCOUNT` ranges (expired + 13 buckets, up to 24h). One pipelined read per queue per poll. No more `ZRANGEBYSCORE`, no member payloads over the wire, no cap, no sampling.
- I removed the expiring-items gauge. The histogram replaces it, one metric instead of two. `le="0"` is for items past their deadline but still queued. The buckets are cumulative, so every item counts in each bucket it expires in.
- The histogram is a snapshot. Each poll replaces it, so it is not monotonic, `rate()` is meaningless, use `histogram_quantile` per scrape.
- The counts are exact broker queries. Only the `_sum` is estimated from bucket midpoints.
- When a read fails, the snapshot is zeroed, so the metric clears instead of staying stale.
- I also fixed the AI suggestions's minor points: the bucket list is fetched once per poll, not three times.

The optional `ZRANGEBYSCORE LIMIT 0 1` for the exact soonest expiry is not included. It stays optional and off by default, like we discussed. If you want it, I can add it as a small follow-up.

### Testing
- **Coverage:** Unit tests, an integration test with `miniredis`, and a live check with Redis 7 (queue frozen).
- **Results:** 
  - 5 items at `+120`/`+120`/`+600`/`+600`/`+3600s` produced: `le=120000=2`, `le=600000=4`, `le=3600000=5`, `_count=5`.
  - 2 minutes later: `le=0=2` after the `+120s` items expired.

Can you look again when you have time?

0xA8hinav:
i have updated the PR title and description

0xA8hinav:
@jtechapps
The pre-commit CI check was failing on a gosec warning (G115, integer overflow) in the deadline proximity collector, and I have pushed the change.

The two conversions are safe because the counts are clamped to zero or above before they are converted, but gosec cannot see that from outside the function, so it flagged them. I suppressed the warning with a nosec comment and wrote the reason next to the code.

The checks should pass now. Thank you for looking at this again

jtechapps (approved):

PR #401, opened 2026-08-16
feat(metrics): add token throughput counter

### LINKED ISSUE #344: [Metrics] Token Throughput counter (input + output tokens processed)
Split out of #217 (comprehensive metrics), which is now closed with the core suite delivered. This is a post-v0.9.0 fast-follow.

### What
A counter of tokens processed by successfully-dispatched requests — the "unit of value" for LLM workloads (cost/throughput dashboards).

- Metric: `async_tokens_total` (counter), labeled by pool plus `direction="input"|"output"` (and the usual `queue_id`/`queue_name` where applicable).

### Approach
In the worker success path (2xx from the IGW), best-effort parse the OpenAI `usage` object from the response body (`prompt_tokens`, `completion_tokens`) and increment the counter. No-op when `usage` is absent or the body isn't parseable.

### Caveats
- Introduces response-body parsing; the processor is otherwise payload-agnostic, so this is **best-effort** and only works for OpenAI-style responses.
- Streaming responses only carry `usage` in the final chunk (and only when `stream_options.include_usage` is set); async/batch traffic is typically non-streaming, so the common path is simple.

0xA8hinav:
## What does this PR do?

Adds an `llm_d_async_async_tokens_total` counter labeled `direction` (`input`/`output`) reporting tokens processed by successfully-dispatched requests, parsed best-effort from the OpenAI usage object in 2xx response bodies (`prompt_tokens` → `input`, `completion_tokens` → `output`).

## Why is this change needed?

Fixes #344. Today the async processor gives no visibility into token throughput, so capacity planning and token accounting are impossible.

## How was this tested?

- Unit tests added/updated (metric helper, usage parser incl. negative clamping, worker dispatch: full/prompt-only/no-usage/3xx)
- Integration/e2e tests added/updated
- Manual testing performed — live Redis + fake IGW: 3 requests × 10 input / 5 output → `async_tokens_total{direction="input"} 30`, `async_tokens_total{direction="output"} 15`

## Demo

<details>
<summary>Screenshot: live output</summary>

Screenshot_2026-08-16_20-06-20

</details>

**Live verification:** Redis 7 + fake IGW, 3 requests × 10 input / 5 output tokens →  
`async_tokens_total{direction="input"} 30`, `{direction="output"} 15`, pool `default`, queue `demo-queue`.

## Checklist

- [x] Commits are signed off (`git commit -s`) per DCO (`PR_SIGNOFF.md`)
- [x] Code follows project contributing guidelines (`CONTRIBUTING.md`)

## Related Issues

Fixes #344

## Release note

```release-note
Token throughput counter `llm_d_async_async_tokens_total` (input/output), parsed best-effort from OpenAI usage in 2xx responses.

### CODE REVIEW

File: pkg/asyncworker/usage.go [hunk 1]
```diff
@@ -0,0 +1,26 @@
+package asyncworker
+
+import "encoding/json"
+
+// ParseUsage extracts the OpenAI usage object from a response body. ok is
```
jtechapps, line 9: Refactor this so that ParseUsage also takes in the requestURL and delegates to a new OpenAI parser if the URL is `/v1/completions` or `/chat/completions`. This is the pattern that llm-d-router follows.

This will allow two things:
1. avoid expensive Unmarshal operation when usage cannot be extracted
2. adds the ability easily extend support to other inference formats.

Abhinav-kodes: Thanks for pointing me to the router. I pulled the repo and checked how it works. I used the same function, `MatchPathSuffix`, from `pkg/epp/framework/common/request/headers.go`. I kept the same suffixes (`v1/completions`, `v1/chat/completions`) and the same slash normalization. The only difference is that I applied it to the worker's full request URL instead of the router's path from headers. So custom base paths and our per-request endpoint overrides still work, and non-OpenAI endpoints now skip the parse. Tests cover all of those cases.
About the delegation: I looked at `ParserRegistry` with the `Parser` interface and `Resolve`. It makes a lot of sense in the router because it handles four formats already. Here, we only have one parser and one place that calls it, so I didn't want to add the whole registry for that. The code is structured so the endpoint check comes first and then the parse.
If we need a second format later, the registry can go in easily. Happy to add it now if you prefer.

0xA8hinav:
@jtechapps
You can re-run the ci, it should be green now

jtechapps (approved):

PR #407, opened 2026-08-25
feat(redis): hot reload sorted-set queue configuration

todayim:
## What does this PR do?

Adds opt-in hot reload for the legacy Redis sorted-set queue-list file through `--redis.ss.queues-config-watch-interval`.

- Periodically re-reads `--redis.ss.queues-config-file`, fingerprints the content, and applies only changed generations.
- Adds a runtime queue registry to `RedisSortedSetFlow`: unchanged queues retain their channel, gate, and consumer; added or modified queues start consuming without a restart; removed queues stop cleanly and leave their Redis backlog intact.
- Extends random-robin and tier-priority merge policies with dynamic source fan-in while keeping each worker pool merged channel stable.
- Preserves the result destination and TTL captured when a request leaves Redis, so a concurrent config change cannot reroute an in-flight result.
- Retries partially applied fan-in updates, ties the watcher to the runner lifecycle, validates the feature before readiness, and removes stale queue snapshot metrics.
- Exposes reload success/error counts and the timestamp of the last successful apply.

The watcher is disabled by default. This first version intentionally targets the existing `--redis.ss.queues-config-file` queue-list surface; whole `--transport-config-file` reload is out of scope.

## Why is this change needed?

Redis sorted-set queue definitions are currently fixed at process startup. Adding, removing, or changing a backlog queue requires restarting the async processor, interrupting consumption and making queue routing updates operationally expensive.

Hot reload lets operators change the active queue set while preserving unchanged workers, queued Redis data, in-flight result routing, and worker-pool dispatch channels. Invalid files and failed applies keep the last good generation active and are retried.

## How was this tested?

- [x] Unit tests added/updated (API routing, merge policies, metrics, Redis reconfiguration, watcher lifecycle/retry, backlog metric cleanup)
- [x] Relevant package suites pass: `go test ./api ./pkg/server ./pkg/redis ./pkg/metrics ./pkg/async/mergepolicy/...`
- [x] Targeted race tests pass for Redis reconfiguration, watcher reload, backlog cleanup, and both dynamic merge policies
- [x] `go vet` passes for all affected packages
- [x] `golangci-lint` reports 0 issues

## Checklist

- [x] Tests pass locally
- [x] Linters pass (`make lint` equivalent)

## Related Issues

None found. This is an independent operational enhancement.

## Release note

```release-note
Redis sorted-set queue-list configurations can now be reloaded at runtime with `--redis.ss.queues-config-watch-interval`, preserving unchanged consumers, in-flight result routing, and Redis backlog while adding, updating, or removing queues.
```

### DIFF
--- a/pkg/async/mergepolicy/randomrobin/random_robin_policy.go
+++ b/pkg/async/mergepolicy/randomrobin/random_robin_policy.go
@@ -98,6 +98,8 @@ func newPoolFanIn(buffer int) *poolFanIn {
 }
 
 func (f *poolFanIn) addSource(ch pipeline.RequestChannel) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	if _, ok := f.sources[ch.Channel]; !ok {
 		f.sources[ch.Channel] = ch
 		f.notifyChanged()
@@ -105,6 +107,8 @@ func (f *poolFanIn) addSource(ch pipeline.RequestChannel) {
 }
 
 func (f *poolFanIn) removeSource(c chan *api.InternalRequest) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	if _, ok := f.sources[c]; ok {
 		delete(f.sources, c)
 		f.notifyChanged()
@@ -203,9 +207,7 @@ func (r *RandomRobinPolicy) merge(channels []pipeline.RequestChannel, pools map[
 		if !ok {
 			return fmt.Errorf("worker pool %q not found in pools map", workerPoolID)
 		}
-		f.mu.Lock()
 		f.addSource(ch)
-		f.mu.Unlock()
 	}
 	return nil
 }
--- a/pkg/async/mergepolicy/randomrobin/random_robin_policy.go
+++ b/pkg/async/mergepolicy/randomrobin/random_robin_policy.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"sync"
 
 	"github.com/llm-d/llm-d-async/api"
 	"github.com/llm-d/llm-d-async/pipeline"
@@ -50,15 +51,20 @@ func NewRandomRobinPolicy(name string, cfg Config) *RandomRobinPolicy {
 	return &RandomRobinPolicy{
 		name:     name,
 		fairness: fairness.New(cfg.FairnessHeader, cfg.FairnessAttribute),
+		pools:    make(map[string]*poolFanIn),
 	}
 }
 
 var _ pipeline.RequestMergePolicy = (*RandomRobinPolicy)(nil)
+var _ pipeline.DynamicRequestMergePolicy = (*RandomRobinPolicy)(nil)
 var _ plugins.Plugin = (*RandomRobinPolicy)(nil)
 
 type RandomRobinPolicy struct {
 	name     string
 	fairness fairness.Stamper
+
+	mu    sync.Mutex
+	pools map[string]*poolFanIn
 }
 
 func (r *RandomRobinPolicy) TypedName() plugins.TypedName {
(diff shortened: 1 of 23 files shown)


jtechapps:
FYI `--redis.ss.queues-config-file` is deprecated and will be removed in the future. PR does mention legacy flag though. Any reason not use use the transport-config and just add a note on which fields are hot reloadable?

pkg/async/mergepolicy/randomrobin/random_robin_policy.go
```diff
@@ -68,7 +74,146 @@ func (r *RandomRobinPolicy) TypedName() plugins.TypedName {
...
+			<-f.changed
+			continue
+		}
+		idx, val, ok := reflect.Select(cases)
+		if idx == 0 {
+			// Source set changed; rebuild the cases.
+			continue
+		}
+		if !ok {
+			f.removeSource(metas[idx-1].Channel)
```
jtechapps, line 154: called without mutex.

todayim:
Good point — thanks for flagging. I'll rework this to watch `--transport-config-file` (redis-sortedset) instead of the deprecated legacy flag:

- hot-apply only the `queues` list (add/update/remove queues without restart);
- changes to any non-queue field (url, retry/result queue names, poll_interval_ms, batch_size, enable_tracing, gate settings) will be logged and keep the last-good generation active — those stay startup-time only;
- flag help and docs will explicitly note that only `queues` is hot-reloadable.

Will push the reworked commits shortly.

pkg/server/queues_reload.go
```diff
@@ -0,0 +1,138 @@
...
+
+			result, err := flow.ReconfigureQueues(queues)
+			if err != nil {
+				metrics.RecordQueueConfigReload(false)
+				logger.Error(err, "Failed to apply queues config; keeping last good queues")
+				return
+			}
+			if len(result.Added) > 0 {
+				if err := policy.AddRequestChannels(result.Added, pools); err != nil {
+					// The flow already consumes the new queues; without a
```
jtechapps: can this actually transiently fail? I don't think so because it can only fail if the queue's pool id doesn't exist which is error with config and can't be tried. Are we assuming that operator will fix the config error and pendingAdded will be processed in the future?

todayim:
Addressed the review feedback in commit `eb5afd6`.

- Moved queue hot reload to the canonical `--transport-config-file` surface with `--transport-config-watch-interval`; only `queues` may change at runtime. Non-queue transport changes are rejected and require a restart.
- Added pre-apply validation for queue `worker_pool_id` references. The old `pendingAdded` retry loop was removed, so a bad config does not wedge the watcher and a corrected file is picked up on the next tick.
- Made queue reconfiguration atomic across the flow and merge policy using a pre-commit callback. A policy registration error leaves the live flow unchanged.
- Protected random-robin fan-in source updates with the same mutex.
- Fixed shutdown ordering so consumption stops before waiting for the reload watcher, avoiding a reload/shutdown wait cycle.
- Added regression coverage for config recovery, empty-queue reloads, policy commit failures, random-robin close/add concurrency, and shutdown ordering.

Validation passed:
`GOWORK=off go test -race -count=1 ./pkg/async/mergepolicy/...`
`GOWORK=off go test -race -count=1 ./pkg/server/...`
`GOWORK=off go test -race -count=1 ./pkg/redis/...`
`GOWORK=off go vet ./pkg/async/mergepolicy/... ./pkg/server/... ./pkg/redis/...`

pkg/async/mergepolicy/randomrobin/random_robin_policy.go
todayim: Fixed in `eb5afd6`: `poolFanIn.addSource` and `removeSource` now both guard `sources` with the same mutex, and the caller-side lock was removed to avoid double locking. Added a race regression test covering concurrent source close and source addition.

pkg/server/queues_reload.go
todayim: Fixed in `eb5afd6`. You are right that an unknown worker-pool reference is a configuration error, not a transient failure. The `pendingAdded` retry loop was removed. Queue pool references are validated before `ReconfigureQueues`; invalid configs keep the last-good state, and the watcher can read and apply a corrected file on the next tick. The policy registration is now a pre-commit callback, so a registration error cannot leave the flow partially updated.

jtechapps:
@todayim can you rebase.

todayim:
Rebased onto the latest  () and force-pushed the signed history. The README conflict was resolved by keeping the upstream configuration reference and adding the queue hot-reload flag to the canonical transport options table. All PR commits remain GPG-signed.

Validation after rebase:
- ok  	github.com/llm-d/llm-d-async/pkg/async/mergepolicy	1.044s
ok  	github.com/llm-d/llm-d-async/pkg/async/mergepolicy/internal/fairness	1.025s
ok  	github.com/llm-d/llm-d-async/pkg/async/mergepolicy/randomrobin	1.352s
ok  	github.com/llm-d/llm-d-async/pkg/async/mergepolicy/tierpriority	1.353s passed
- ok  	github.com/llm-d/llm-d-async/pkg/server	1.846s passed on rerun (one initial run hit the existing timing-sensitive backlog snapshot test)
- ok  	github.com/llm-d/llm-d-async/pkg/redis	22.507s passed
-  passed

jtechapps:
@todayim there seems to be an issue with the merge policies that is failing the CI.

todayim:
The CI failure was caused by the existing integration test expecting the merged channel to close after all source channels close. With dynamic fan-in, the merged channel intentionally remains open so queues can be added later, so that expectation is no longer valid.

Fixed in commit `7708adb`: the test now waits for all expected messages (100 total) and verifies delivery exactly once, without waiting for the merged channel to close.

Local validation passed:
- `GOWORK=off go test -tags=integration -count=1 ./test/integration -run TestRandomRobinPolicy_ConcurrentProducers -v`
- `GOWORK=off go test -race -count=1 ./pkg/async/mergepolicy/... ./pkg/server/... ./pkg/redis/...`

jtechapps (approved):

PR #424, opened 2026-09-02
feat: enforce leased Redis dispatch rates

sachalmalick:
This PR adds a Redis-backed gate that enforces Planner's leased, pool-wide dispatch-attempt rate across Async replicas. It provides a fail-closed control path for dynamic drain limits while keeping an explicit zero-rate lease distinct from an invalid command.

CLOSES: LLM-180

### Review Stack

Review bottom-to-top: **#425 → #426 → #424** (this PR is layer 3 of 3). The child PRs remain based on `main` because the upstream repository cannot select fork-only head branches as PR bases and native stacked PRs are disabled for this repository.

### How This Was Implemented

- Registered a `redis-leased-rate` gate with validated Redis address, pool, control key, state key, and burst-window configuration.
- Validated the `llm-d.ai/v1alpha1` command fields and rejected missing, malformed, mismatched, expired, or unreadable leases.
- Enforced a shared token bucket atomically in Lua using Redis time, with bounded burst capacity and lease-aligned state expiry.
- Updated drain-limit telemetry on each gate evaluation and refused dispatch safely on Redis or control-plane failures.
- Documented pool-level `wait-on-refuse` wiring, command publication, pause semantics, and Valkey compatibility.

<details>
<summary>Walkthrough</summary>

- `pkg/async/inference/flowcontrol/gate_factory.go` constructs the new gate and shares Redis clients by address.
- `pkg/redis/leased_rate_gate.go` reads leases, evaluates budget state, acquires tokens, and emits Planner-facing metrics.
- The embedded Lua script gives all Async replicas one aggregate rate ceiling and uses Redis time to avoid host-clock skew.
- Unit and integration fixtures cover configuration, fail-closed behavior, explicit pause, token refill, skew, shared limits, and wait/requeue behavior.

</details>

### Validation

- `git diff --check origin/main..HEAD` and `gofmt` over changed Go files — pass.
- Non-E2E root unit packages and `go test ./api/... ./pipeline/... ./producer/...` — pass.
- `go test -tags=integration ./test/integration/` — pass.
- Repository `golangci-lint` under Go 1.26 — pass (`0 issues`) on the complete three-layer stack.
- `helm lint charts/llm-d-async` and rendered templates for the default five-minute, custom 90-second, and disabled (`0`) gate-wait settings — pass.
- GitHub `DCO` and `signed-commits` checks for `63f14c5` — pass; GitHub reports a valid signature and matching author, committer, and sign-off identity.


api/dispatch_control.go
```diff
@@ -0,0 +1,50 @@
+package api
+
+import (
+	"fmt"
```
jtechapps, line 4: file formatting is weird

jtechapps:
Needs rebase and signed commits.

api/dispatch_control.go
sachalmalick: [AI] Fixed in e6a022d. I replayed `api/dispatch_control.go` onto current `main` and reran `gofmt`; the changed-file `gofmt -l` check returned no paths, and byte inspection shows normal LF text. I kept the standard grouped import block because it is Go's canonical formatting.

sachalmalick:
[AI] Fixed in 63f14c5. I replayed this layer as one commit after #426 while retaining upstream durable Redis dequeue behavior. Please review #425 → #426 → #424; this PR remains main-based because GitHub cannot use the fork-only #426 head as an upstream base and native stacked PRs are disabled here.

The author and committer are now `Sachal Malick <ssrivastavam@nvidia.com>`, the commit has a matching `Signed-off-by` trailer, and GitHub reports its signature valid; both `signed-commits` and `DCO` pass. Repository lint under Go 1.26 also passes with `0 issues`.

jtechapps:
**`Budget` validates the lease against the host clock while `Apply` enforces against the Redis clock**

`pkg/redis/leased_rate_gate.go`

`Budget` reads the lease and validates its expiry using the host wall clock:

```go
func (g *RedisLeasedRateGate) Budget(ctx context.Context) float64 {
    limit, err := g.readLimit(ctx, g.now())   // g.now == time.Now (host clock)
    ...
}
// readLimit -> limit.ValidateAt(now)  // checks valid_until against host time
```

`Apply`, however, enforces the same lease against the **Redis** clock inside the Lua script:

```lua
local redis_time = redis.call("TIME")
...
local valid_until_invalid = ... or valid_until <= now   -- Redis time
```

The stated design of the Lua path is to make Redis the single authoritative clock so per-replica host-clock skew can't affect enforcement (and there's a test named `TestRedisLeasedRateGateUsesRedisTimeAcrossSkewedInstances` asserting exactly that). But `Budget` bypasses Redis time and uses the local host clock, so the two disagree near lease expiry:

- **Host clock ahead of Redis:** `Budget` reports the lease expired — returns `0` and publishes `drain_limit_lease_valid=0` via `SetDrainLimit` — while `Apply`/Redis still consider it valid and keep admitting tokens.
- **Host clock behind Redis:** `Budget` reports valid/`rate>0` after Redis already treats the lease as expired.

The divergence window is proportional to the host-vs-Redis clock skew. Any `Budget` consumer (composite gate, pubsub batch-sizing) and the drain-limit telemetry can then flap or disagree with what `Apply` actually enforces.

**Suggested fix:** have `Budget` obtain "now" from Redis (`TIME`) as `Apply` does, so both validate against the same clock — or, if `Budget` is intended to be advisory/best-effort, document that it uses host time and may briefly disagree with enforcement near expiry.

**Related test gap:** `TestRedisLeasedRateGateUsesRedisTimeAcrossSkewedInstances` sets `gate.now`/`other.now` but never calls `Budget`, so it doesn't exercise the host-clock path — precisely the path that diverges from Redis time. The test's name oversells what it verifies.

sachalmalick:
[AI] Fixed in 0726b87. Rebased the stack onto current main and changed Budget to read the lease plus Redis TIME in one atomic MULTI/EXEC snapshot, preserving fail-closed behavior. A two-direction host/Redis clock-skew regression now proves Budget telemetry agrees with Apply. The refreshed commits are signed and DCO-clean.

jtechapps (approved):

Issue #151, opened 2026-05-05
Producer: remove GetResultWithTimeout in favor of context-based timeout


yizhaodev:
`GetResultWithTimeout()` is redundant — callers can achieve the same behavior with `context.WithTimeout` + `GetResult()`.

Removing it simplifies the API surface without losing functionality.

PR #434, opened 2026-09-05
fix: honor context cancellation while waiting for results

git-jxj:
## What does this PR do?

Use one-second blocking Redis reads in `GetResult`, checking the caller's context between empty reads. Results received by a read are still returned to the caller.

## Why is this change needed?

`GetResult` promises to stop when its context is cancelled, but `BRPOP` with a zero timeout blocks indefinitely on an empty queue. Both a deadline and cancellation after the read starts leave callers waiting until a result arrives. The bounded reads allow cancellation to be observed after approximately one second without leaving a background consumer behind.

## How was this tested?

- [x] Unit tests added/updated: cancellation during a read, deadline expiry, subsequent reads on the same producer, and results arriving after an empty poll.
- [x] Manual testing performed: Redis 6.0.16 reproduced both hangs before the fix; afterwards both calls returned their context errors in about one second, and a result published after 1.3 seconds was delivered normally.

Validation: `make test-all` (with `NO_PROXY=*` for the local IPv6 health endpoint), producer `go test -race ./...`, pipeline `go test ./...`, and all pre-commit hooks. The root-module lint hook and producer-module lint with `--new-from-rev HEAD` pass. A full producer-module lint run also has pre-existing findings on unmodified upstream.

## Checklist

## Related Issues

Related to #151.

## Release note

```release-note
Redis producer GetResult now observes context cancellation while waiting on an empty result queue, with approximately one second of cancellation latency.
```

### DIFF
--- a/producer/redis_sortedset_producer.go
+++ b/producer/redis_sortedset_producer.go
@@ -257,22 +257,31 @@ func newRequestToken() (string, error) {
 }
 
 // GetResult retrieves a result from the Redis list, blocking until one is available.
+// Cancellation is checked between one-second blocking waits, so an empty queue
+// can delay observing cancellation by approximately one second.
 func (p *RedisSortedSetProducer) GetResult(ctx context.Context) (*api.ResultMessage, error) {
-	// Use BRPOP (blocking right pop) to wait for a result
-	result, err := p.client.BRPop(ctx, 0, p.resultQueueName).Result()
-	if err != nil {
-		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-			return nil, err
+	// A zero BRPOP timeout blocks indefinitely: go-redis does not interrupt an
+	// in-flight read when the context is cancelled. Bound each empty wait so we
+	// can check the context without leaving a background pop consuming results.
+	for {
+		if err := ctx.Err(); err != nil {
+			return nil, fmt.Errorf("failed to get result: %w", err)
 		}
-		return nil, fmt.Errorf("failed to get result: %w", err)
-	}
+		result, err := p.client.BRPop(ctx, time.Second, p.resultQueueName).Result()
+		if errors.Is(err, redis.Nil) {
+			continue
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to get result: %w", err)
+		}
 
-	// BRPOP returns [queueName, value]
-	if len(result) != 2 {
-		return nil, errors.New("unexpected BRPOP result format")
-	}
+		// BRPOP returns [queueName, value]
+		if len(result) != 2 {
+			return nil, errors.New("unexpected BRPOP result format")
+		}
 
-	return p.parseResult(result[1])
+		return p.parseResult(result[1])
+	}
 }
 
 // parseResult parses a JSON result message.
--- a/producer/redis_sortedset_producer_test.go
+++ b/producer/redis_sortedset_producer_test.go
@@ -312,6 +312,75 @@ func TestGetResultWithContextTimeout(t *testing.T) {
 	})
 }
 
+func TestGetResultWhileWaitingContextEnds(t *testing.T) {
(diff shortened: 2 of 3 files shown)


git-jxj:
The GitHub Actions runs for current head `c5a3945c74f3` are waiting for maintainer approval (`action_required`), including [DCO Sign-off Check](https://github.com/llm-d/llm-d-async/actions/runs/33968199844), [Pre-commit](https://github.com/llm-d/llm-d-async/actions/runs/33968199859) (and 2 other workflows). Could a maintainer approve the pending workflow runs so CI can validate this head?

producer/redis_sortedset_producer.go
```diff
@@ -257,22 +257,31 @@ func newRequestToken() (string, error) {
...
 // GetResult retrieves a result from the Redis list, blocking until one is available.
+// Cancellation is checked between one-second blocking waits, so an empty queue
+// can delay observing cancellation by approximately one second.
 func (p *RedisSortedSetProducer) GetResult(ctx context.Context) (*api.ResultMessage, error) {
-	// Use BRPOP (blocking right pop) to wait for a result
-	result, err := p.client.BRPop(ctx, 0, p.resultQueueName).Result()
-	if err != nil {
-		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-			return nil, err
+	// A zero BRPOP timeout blocks indefinitely: go-redis does not interrupt an
```
jtechapps, line 263: Why is this change and loop needed? Has an issue been observed. The ctx was already passed to `p.client.BRPop(ctx, 0, p.resultQueueName).Result()` does the Redis client not respect the context cancelation?

jtechapps: oh I see there is bug open on redis-go about this issue.

producer/redis_sortedset_producer.go
```diff
@@ -257,22 +257,31 @@ func newRequestToken() (string, error) {
...
+	// in-flight read when the context is cancelled. Bound each empty wait so we
+	// can check the context without leaving a background pop consuming results.
+	for {
+		if err := ctx.Err(); err != nil {
+			return nil, fmt.Errorf("failed to get result: %w", err)
+		}
+		result, err := p.client.BRPop(ctx, time.Second, p.resultQueueName).Result()
```
jtechapps, line 270: how was this timeout value decided?

producer/redis_sortedset_producer.go
git-jxj: Yes, this addresses the in-flight blocking-read behavior you found. Passing `ctx` alone does not wake an already-blocked `BRPOP` read when cancellation occurs, and a zero Redis timeout can therefore leave an empty-queue wait blocked indefinitely. The loop gives Redis a finite blocking timeout, retries `redis.Nil` while the context remains active, and checks cancellation before the next pop. It also avoids returning while a background pop continues consuming later results.

The regression tests cover cancellation and deadline expiry after the blocking command has started, verify that a subsequent caller can receive a later result, and check that empty polls do not end an active wait. Re-ran `go test ./producer -run 'TestGetResult|TestContextCancellation' -count=1`: passed.

producer/redis_sortedset_producer.go
git-jxj: One second is the smallest positive timeout supported by the `BRPop` implementation in our pinned go-redis v9.22.0: it serializes the timeout with [`formatSec`](https://github.com/redis/go-redis/blob/v9.22.0/commands.go#L41-L51), which converts positive sub-second durations to one second. Zero means block indefinitely.

This keeps an idle waiter to roughly one Redis command per second and lets it observe cancellation after approximately one blocking interval under normal Redis/network conditions. It does not add a one-second delay when a result arrives: Redis returns the blocking pop immediately. The method comment documents this cancellation latency. This is a client-granularity choice, not a value derived from a benchmark.

jtechapps (approved):

PR #437, opened 2026-09-08
feat(producer): add durable Redis result delivery

### LINKED ISSUE #423: [Bug]: Results can be lost after Producer GetResult removes them before consumer persistence
### Contact Details

hannahz@nvidia.com

### What happened?

The Redis sorted-set Producer currently retrieves results with a destructive `BRPOP`: https://github.com/llm-d/llm-d-async/blob/main/producer/redis_sortedset_producer.go#L259-L275

Once `GetResult` returns, the result no longer exists in Redis. If the consumer exits after receiving the result but before durably incorporating it into its own state, the result is permanently lost.

I hit the surrounding failure mode while validating llm-d Batch Gateway with Async. Batch Gateway can route results to the correct healthy Processor replica after llm-d-batch-gateway#662, but the owning Processor still has a crash window:

1. Async durably publishes a terminal result.
2. The Processor calls `GetResult`; `BRPOP` deletes the result from Redis.
3. The Processor exits before checkpointing the terminal record or updating the batch.
4. A replacement Processor has nothing to redeliver, so the batch can remain `in_progress` or `cancelling` indefinitely.

This is separate from request-side durable dequeue in #404/#412. Those changes protect accepted requests until Async durably publishes a result. This issue covers delivery of that result from Async's Producer API to a consumer that can durably accept it.

Expected contract:

- Result delivery is at least once until the consumer explicitly acknowledges durable acceptance.
- A consumer failure before acknowledgement makes the result available for redelivery.
- A stale consumer cannot acknowledge or delete a result owned by a newer delivery.
- A failure after consumer checkpointing but before acknowledgement may redeliver, so consumers deduplicate using the stable internal request generation (`result.ID` plus `Routing.RequestToken`), never the user-provided `custom_id`.
- Exactly one terminal record is exposed externally per accepted request.

Proposed ownership and direction:

Async should own the result transport and acknowledgement contract. Batch Gateway #645 should own its durable job manifest, per-request checkpoint, takeover generation, output reconstruction, and duplicate suppression.

Add durable receive/ack semantics to the Producer result contract rather than implementing transport-specific recovery in Batch Gateway. One possible API shape is a result-delivery handle containing the result, an ownership token, and an explicit `Ack(ctx)`. The Redis implementation could use a claim/lease/ack structure or Redis Streams consumer-group semantics. The specific storage mechanism is less important than these boundaries:

1. Receiving a result must not permanently delete it.
2. Acknowledgement must be owner-fenced and idempotent.
3. Lease expiry or negative acknowledgement must make the result eligible for redelivery.
4. The API must continue to honor per-request or per-replica `ResultQueueName` routing.
5. Success, HTTP error, non-HTTP error, deadline, and cancellation results must all use the same durable delivery path.

Acceptance criteria:

- Hard-kill before result acknowledgement redelivers the result.
- Hard-kill after consumer checkpoint but before acknowledgement may redeliver, and the composed consumer test produces no duplicate output.
- A stale acknowledgement cannot remove a newer delivery.
- Result routing remains compatible with replica-specific queues from llm-d-batch-gateway#662.
- Recovery time and lease/reclaim configuration are observable and documented.
- Mixed-version rollout behavior is documented.

### Version

`main` as of 2026-09-01. The destructive `BRPOP` behavior is also present in `v0.9.0`.

### Steps to Reproduce

1. Configure the Redis sorted-set Producer with a result queue.
2. Publish a valid `ResultMessage` to that queue.
3. Call `Producer.GetResult` and verify that the result is returned.
4. Terminate the consumer before it persists or otherwise accepts the result.
5. Start a replacement consumer and call `GetResult`.
6. Observe that the queue is empty and the previously returned result cannot be recovered.

### Environment

```markdown

```

### Relevant log output

```shell

```

hhzhang16:
## What does this PR do?

- add lease-based durable result receive, renewal, and owner-fenced ACK
- retain request tokens in a backward-compatible Redis result envelope
- cover expiry redelivery, fencing, route isolation, terminal result shapes, and cleanup
- document consumer checkpointing and rollout constraints

## Why is this change needed?

Bug fix for #423 

## How was this tested?

## Checklist

## Related Issues

Closes #423 

## Release note

```release-note
Add durable Redis result delivery with lease-based redelivery and acknowledgement.
```


jtechapps:
**Durable results can be silently lost, poison records stall the consumer, and the poll loop wastes entropy**

Three issues in `ReceiveResult` (`producer/redis_result_delivery.go`):

**1. A claimed-but-unparsable result is permanently deleted (`:226-234`)**

When `receiveResultScript` returns `{1, payload, claimID}`, the record has already been `RPOP`'d from `pending` and moved into the `claimed`/`owners`/`idx` claim state — the lease is now the only thing keeping it durable. If `parseResult(payload)` then fails, we run `discardResultDeliveryScript`, which deletes it from all three claim structures. At that point the record is gone from `pending` *and* claim state — never acked, never redelivered.

This can fire even though Lua already validated the payload, because `cjson.decode` accepts shapes that Go's `json.Unmarshal` rejects (e.g. a `status_code` typed as string, a numeric overflow, a field shape the struct tags don't accept). A schema skew during a rolling upgrade is the realistic trigger, and it contradicts the "durable until AckResult" contract with no trace left behind.

**2. Poison records return an error indistinguishable from a real Redis failure (`:216-217`)**

The `status == -1` path returns a plain `errors.New(...)`, which looks identical to the genuine-failure path at line 194 (`failed to receive durable result: %w`). A normal consumer loop —

```go
for {
    d, err := p.ReceiveResult(ctx)
    if err != nil { break } // or back off
    process(d)
}
```

— will exit or back off on a single poison message, even though the record is already gone from `pending` and the next call would return the next valid result. One bad message stalls the whole route.

**3. `newRequestToken()` (crypto/rand) runs on every poll tick (`:181`)**

`ownerToken` is generated inside the `for` loop but only consumed on a successful claim (it's returned in the `ResultDelivery` and never reused after). On every empty poll, every all-tombstoned poll (`status == 0` → `continue`), and every poison record, a fresh token is minted from `crypto/rand` and discarded. With the small reclaim intervals in the tests (1ms/5ms) and a consumer blocked on an empty route, that's continuous `crypto/rand.Read` + allocation purely while waiting. Hoisting the token above the loop is safe (each `ReceiveResult` call still gets its own) and strictly cheaper.

**Suggested direction**

- **2:** return a distinguished sentinel (e.g. `ErrUnparseableResult`) from both the `-1` path and the `parseResult`-failure path, so callers can `errors.Is(...)` and `continue` instead of treating it as fatal.
- **1:** decide the policy explicitly. If we want at-least-once durability for malformed results, route them to a dead-letter queue (or at minimum log the full raw payload + `claimID` at error level before discarding, so an operator can recover them). **If we're comfortable dropping unparsable results, the current discard is fine as-is — just make it a deliberate, documented choice paired with the sentinel error above** so consumers can skip-and-continue rather than dropping *and* stalling.
- **3:** move `newRequestToken()` above the `for` loop.

jtechapps (approved):

Issue #417, opened 2026-08-27
[Feature]: Remove Redis PubSub transport


shimib:
### Problem Statement

Suggesting removing the ephemeral Redis PubSub transport. 
Starting by first deprecating it. 

### Proposed Solution

Deprecate the transport 

### Willingness to Contribute

Yes, I can submit a PR

jtechapps:
Reason for deprecating: is that it is ephemeral and not persisted which defeats the value proposition of this component. Also it doesn't support many of the other features for the other transports and people aren't really using it.

PR #442, opened 2026-09-11
feat: deprecate redis-pubsub transport

804533125:
## What
Deprecates the `redis-pubsub` transport (step 1 of #417). No functional change: it still works, but users get a warning pointing at `redis-sortedset`.

- Processor logs a startup warning whenever the effective transport is `redis-pubsub` (including via the default), naming the replacement (`pkg/server`)
- Table-driven unit test for the warning (default / new surface / legacy flag / other transports)
- README and chart `values.yaml` mark `redis-pubsub` as deprecated
- Helm chart prints a migration warning in the install NOTES

## Why
#417 removes the ephemeral Redis PubSub transport, starting with deprecation. Per the N+2 policy, it keeps working through v0.11.x and may be removed no earlier than v0.12.0 (same pattern as #378 / #388).

## Testing
- [x] `TestWarnDeprecatedTransport` (5 cases); `go test ./pkg/server/` passes
- [x] Manual: warning with default config; no warning with `--message-queue-impl=redis-sortedset`
- [x] `gofmt`/`go vet`/`go build` pass across all four modules; `make test` and `make lint` pass

## Checklist
- [x] Commits signed off (DCO) and signed
- [x] Follows CONTRIBUTING.md
- [x] Tests + linters pass
- [x] Docs updated

Related: #417

## Release note

```release-note
**Deprecated:** the `redis-pubsub` transport is deprecated. It still works and will keep working through this release and the next, but it is scheduled for removal (no earlier than `v0.12.0`, per the N+2 deprecation policy). The processor logs a startup warning whenever the effective transport is `redis-pubsub` (including via the default), and the Helm chart prints the same notice in its install NOTES. Prefer `redis-sortedset` (recommended for production) or `gcp-pubsub`. No configuration change is required for existing deployments.
```


jtechapps:
Release-note fragment starts with a UTF-8 BOM — breaks CI (blocking) — release-notes.d/unreleased/442.md:1
  The file begins with EF BB BF (UTF-8 BOM) before ---, so its first line is \ufeff---, not ---. hack/lint-release-notes.sh (run by make ci / make 
  release-notes-lint) checks [[ "$(sed -n '1p' "$f")" != "---" ]] and will fail with "must start with '---' front matter" and exit non-zero.

jtechapps (approved):

PR #452, opened 2026-09-15
Add sql transport backed by Postgres

wseaton:
Adds a `sql` transport: dispatchers lease hash partitions of each queue and batch dispatch, acks, and submits, and each request's payload has its own column, so dispatch decodes only the envelope. Stacked on #458; its first five commits belong to that PR ([report](https://gist.github.com/wseaton/c97e31be9b92cde670b952a09c47faf6)).

End to end: sql vs redis-sortedset
Store throughput
Result batch size vs throughput

## Release note

```release-note
Adds a `sql` transport backed by Postgres. Dispatchers lease hash partitions of each queue and batch dispatch, acks, and cancellation checks, and each request's envelope and payload are stored in separate columns.
```


hack/transport-bench/main.go
```diff
@@ -0,0 +1,386 @@
...
+
+func (c sqlClient) results(ctx context.Context) ([]*api.ResultMessage, error) {
+	res, err := c.p.GetResults(ctx, 64)
+	if err != nil {
+		return res, fmt.Errorf("results: %w", err)
+	}
+	return res, nil
+}
+
+func (c sqlClient) close() { _ = c.p.Close() }
```
hexfusion: question: Close() is idempotent?

producer-sql/sqlqueue/store.go
```diff
@@ -0,0 +1,722 @@
...
+	if o.tracerProvider != nil {
+		cfg.Tracer = otelpgx.NewTracer(otelpgx.WithTracerProvider(o.tracerProvider))
+	}
+	db := stdlib.OpenDB(*cfg)
+	s := &Store{db: db, tracer: noop.NewTracerProvider().Tracer("")}
+	if o.tracerProvider != nil {
+		s.tracer = o.tracerProvider.Tracer("github.com/llm-d/llm-d-async/producer-sql/sqlqueue")
+	}
+	if err := db.PingContext(ctx); err != nil {
+		_ = db.Close()
```
hexfusion: should we consider logging this as debug, I am not 100% sure the consequence of close failure in this context but I think we would want to know.

producer-sql/sqlqueue/store.go
```diff
@@ -0,0 +1,722 @@
...
+
+func (s *Store) Ping(ctx context.Context) error {
+	if err := s.db.PingContext(ctx); err != nil {
+		return fmt.Errorf("sqlqueue: ping: %w", err)
+	}
+	return nil
+}
+
+func (s *Store) Close() error {
+	if err := s.db.Close(); err != nil {
```
hexfusion: question: is this idempotent?

